PR #5, opened 2021-02-24
Add support to Locale Date String

aperezl:
remove tinytime library and use toLocaleDateString to add internationalization


timlrx:
Thanks for the contribution!

PR #9, opened 2021-03-20
Add feature and fix a bug

CypressKuo:
feature:
If there's no social link in siteMetadata.json, it can hide icons automatically.

bug:
useTheme is not working in the first render.

### DIFF
--- a/components/ThemeSwitch.js
+++ b/components/ThemeSwitch.js
@@ -1,7 +1,12 @@
+import { useEffect, useState } from 'react'
 import { useTheme } from 'next-themes'
 
 const ThemeSwitch = () => {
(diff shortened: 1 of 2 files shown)


DoctorDerek:
I had to make a similar change to the dark theme (using useEffect) when I wanted to resolve the system theme, but I think it worked fine to remember the theme and otherwise function as is. That wasn't your experience @CypressKuo

timlrx:
Thanks for the fix. This is related to https://github.com/pacocoursey/next-themes#avoid-hydration-mismatch where the button state displays incorrectly on first render if using dark theme (moon icon instead of sun).

PR #12, opened 2021-03-31
Remove experimental flag modern: true

DoctorDerek:
This feature got dropped by Next.js so should be removed, see: https://github.com/vercel/next.js/pull/19275

Arguably, it could still be worth keeping (people mention 3-6% bundle size improvement, 21% bundle size improvement when using Preact), but it's a bit of a landmine.

I haven't actually compared bundle sizes or looked at the effect of this change. What do you think @timlrx?

As discussed in that pull request over at Next.js, it's incompatible for when they migrate to Webpack 5. I think removing this feature is currently in the /canary version of Next.js, not a released version.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -4,9 +4,6 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 
 module.exports = withBundleAnalyzer({
   pageExtensions: ['js', 'jsx', 'md', 'mdx'],
-  experimental: {
-    modern: true,
-  },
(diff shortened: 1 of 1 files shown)


timlrx:
Thanks, I see that next.js 10.1 has upgraded to webpack 5. You probably encountered an error since you mentioned in a previous post that you deleted package-lock.json and re-installed it. I Will merge this together will the next version upgrade.

timlrx:
updated in #14 and use webpack 5

PR #13, opened 2021-04-01
Update react & react-dom from 16.13.1 to ^17.0.2

DoctorDerek:
This would update react & react-dom from 16.13.1 to ^17.0.2

I'm happy to go through and remove all the now-unnecessary `import "react" from React` statements if you'd like.

There basically aren't any [breaking changes in React 17](https://javascript.plainenglish.io/the-biggest-change-in-react-17-that-no-one-is-talking-about-b62905beb2d2), so this should be a pretty safe change.

image


timlrx:
Any reason for the version upgrade? We are using preact for the production version anyway. I guess we can upgrade all the dependencies at one go and create a new tag.

timlrx:
upgraded to react 17 in #14!

PR #14, opened 2021-04-02
update dependencies

timlrx:
Update to next.js v10.1, webpack 5 and use @tailwindcss/jit


DoctorDerek:
Huh, I thought prose was broken with jit, but it seems to work fine in your new build.

The React upgrade to 17 suppresses the error message about Next 11 (from npm run dev).

It should permit removing `import React from 'React'` statements from the codebase (though I'm not sure if Preact in production changes that at all).

PR #18, opened 2021-04-06
Upgrade Tailwind to 2.1.1 keeping jit (Just in Time) engine enabled per https://blog.tailwindcss.com/tailwindcss-2-1

DoctorDerek:
Nothing as exciting in tech as upgrading to a much better solution and then immediately having to reconfigure said solution in a different way for no particular reason, right @timlrx ? Haha


timlrx:
yup, that's what I thought too when I saw the announcement haha

PR #19, opened 2021-04-06
Create .gitattributes

DoctorDerek:
> There once was a man from Nantucket,
> who developed on a Macintosh bucket.
> He set up his ESLint to flag Prettier errors,
> And Windows developers got CR* terrors.
> The man was no fan, so he hatched up a plan:
> He would ask Git to enforce LF** characters.
> So now developers rejoice, because of his choice:
> They can keep their Windows kit, not chuck it.

*Carriage Return, **Line Feed

Original poetry! Boom 💥

`eslintrc.js`
```
module.exports = {
  ...
  extends: [
    'plugin:prettier/recommended',
  ],
  rules: {
    'prettier/prettier': 'error',
  }
}
```

See discussion https://github.com/tailwindlabs/tailwindcss/pull/3760

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*		text eol=lf


timlrx:
Nice poetry :)

I configured my vscode end of line sequence but this should work more generally.

DoctorDerek:
Thanks @timlrx 

-----------

Just to ask a follow-up question, configuring the VS Code end of line sequence doesn't actually change anything, right?
image

Like, I'll still pull down the entire repository on Windows using CR/LF endings, and I'd have to open and save each file to change the endings. (Prettier handling line endings is a pretty new feature, so if Prettier doesn't catch it then I'd have to toggle them manually in the VS Code Status Bar or have a .editorconfig file to format them using that tool.)

--------

To wit:
image

still produces
image

And even updating the Prettier config isn't doing it for me:
image
My guess is that setting may not be supported by the Prettier VS Code extension, since it's newish.

----------

In a quick test, I found that in TailwindCSS's repo (which comically lacks a Prettier config file, and they haven't accepted my PR), the `.editorconfig` file does in fact force LF endings (but only on save), independent of that VS Code setting.
image

That VS Code setting you mentioned only affects new files I create. It doesn't do anything to existing files, or the repository as a whole. And even if I used that, I'd still pull down CR/LF line endings on Windows the next time I clone that repository, since that's GitHub's default setting.

Without `.gitattributes`, the only solution to suppress the "prettier/prettier" error is running ESLint on the repository (as would happen automatically with husky when I commit from the command line) -- or changing the ESLint configuration.

------------

So just to recap, the VS Code setting wouldn't seem to have the least bit to do with fixing this particular problem affecting any Windows developer cloning a repository without a `.gitattributes` file that uses the "prettier/prettier" ESLint rule. Am I understanding that correctly?

timlrx:
Yes, I think you are right. Since I have edited most files they all seem fine to me but to someone who has cloned the repo you will see a lot of red squiggly lines

DoctorDerek:
FWIW I was trying different configurations, and `.editorconfig` will force files to be LF on save (assuming that the "LF" end of line option is set in the `.editorconfig` file and that EditorConfig is configured and running in VS Code or another editor).

I was wondering why I was still having to manually change new files from CRLF->LF in VS Code in one repo but not another, and that's the difference. 😀

`https://github.com/tailwindlabs/tailwindcss/blob/master/.editorconfig`
```
root = true

[*]
indent_style = space
indent_size = 2
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
```

PR #23, opened 2021-04-12
Update husky (^4.3.7) to latest (^6.0.0)

### LINKED ISSUE #11: Remove husky as a requirement?
Hey @timlrx ,
Do you think it's better to include instructions on using husky or to remove it as a dependency?

I went to commit locally after cloning the template, but I couldn't because husky not being configured threw an error.

That meant I had to pause what I was doing to `npm uninstall husky` -- then commit the package.json & package-lock.json files -- and then commit my actual changes.

I'm happy to update your template for whichever solution you think is best.

DoctorDerek:
This would close #11 for Windows developers using GitHub Desktop.

(In other words, I can commit normally & the husky hooks will run -- which wasn't true with ^4.3.7 even after npm install.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2251,12 +2251,6 @@
         "readdirp": "~3.5.0"
       }
     },
-    "ci-info": {
-      "version": "2.0.0",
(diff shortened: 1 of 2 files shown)


timlrx:
Great!

Issue #21, opened 2021-04-09
Errors on Vercel build


theapplegates:
I am getting some strange errors when I try to import the site to Vercel. I use the default values. I had the site up until a few days ago when it would fail to build. I am not a developer am not sure where to start. Here is the Vercel log file.

I made a gist for a more readable format. Not sure why it looks bad in a preview.

https://gist.github.com/theapplegates/e3d2d065ec51b8531ce13631cfaaf0e4

`01:17:37	Cloning github.com/theapplegates/crispy-parakeet-tailwind (Branch: main, Commit: 0e615ee)
01:17:38	Cloning completed: 652.690ms
01:17:38	Analyzing source code...
01:17:40	Installing build runtime...
01:17:45	Build runtime installed: 5222.614ms
01:17:49	Looking up build cache...
01:17:49	Build cache not found
01:17:51	Installing dependencies...
01:17:51	yarn install v1.22.10
01:17:51	warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
01:17:51	[1/4] Resolving packages...
01:17:51	[2/4] Fetching packages...
01:18:04	info fsevents@2.3.2: The platform "linux" is incompatible with this module.
01:18:04	info "fsevents@2.3.2" is an optional dependency and failed compatibility check. Excluding it from installation.
01:18:04	[3/4] Linking dependencies...
01:18:04	warning " > next-mdx-remote@2.1.3" has incorrect peer dependency "react@>=16.13.1 <=17.0.1".
01:18:04	warning " > next-mdx-remote@2.1.3" has incorrect peer dependency "react-dom@>=16.13.1 <=17.0.1".
01:18:04	warning " > file-loader@6.2.0" has unmet peer dependency "webpack@^4.0.0 || ^5.0.0".
01:18:15	[4/4] Building fresh packages...
01:18:16	Done in 24.48s.
01:18:16	Detected Next.js version: 10.1.3
01:18:16	Running "yarn run build"
01:18:16	yarn run v1.22.10
01:18:16	$ next build && node ./scripts/generate-sitemap
01:18:17	info  - Using webpack 5. Reason: future.webpack5 option enabled https://nextjs.org/docs/messages/webpack5
01:18:17	info  - Checking validity of types...
[...]

PR #24, opened 2021-04-15
Update .gitattributes

DoctorDerek:
> There once was a man from Nantucket,
> who developed on a Macintosh bucket.
> He set up his ESLint to flag Prettier errors,
> And Windows developers got CR* terrors.
> The man was no fan, so he hatched up a plan:
> He would ask Git to enforce LF** characters.
> So now developers rejoice, because of his choice:
> They can keep their Windows kit, not chuck it.

*Carriage Return, **Line Feed

Original poetry! Boom 💥

`eslintrc.js`
```
module.exports = {
  ...
  extends: [
    'plugin:prettier/recommended',
  ],
  rules: {
    'prettier/prettier': 'error',
  }
}
```

See discussion https://github.com/tailwindlabs/tailwindcss/pull/3760

---

# New `.gitattributes` file

I updated the `.gitattributes` file to explicitly avoid changing binary files, based on best practices from https://gitattributes.io/

```
## Source: https://github.com/alexkaratarakis/gitattributes
## Modified * text=auto to * text eol=lf to force LF endings.

## GITATTRIBUTES FOR WEB PROJECTS
#
# These settings are for any web project.
#
# Details per file setting:
#   text    These files should be normalized (i.e. convert CRLF to LF).
#   binary  These files are binary and should be left untouched.
#
# Note that binary is a macro for -text -diff.
######################################################################

# Auto detect
##   Force LF line endings automatically for files detected as
##   text and leave all files detected as binary untouched.
##   This will handle all files NOT defined below.
*                 text eol=lf
```

The new `.gitattributes` file continues on for about 200 lines total while achieving the same result:

*Windows developers will have LF line endings for text files, meaning they will not see ESLint "prettier/prettier" errors.*

This improved `.gitattributes` file prevents corrupting binary files (e.g. `.png` or `.woff`), which can happen if git finds a false "CRLF" inside their binary code. Just ask my Vercel deploy about that 😄

### Comparison with previous `.gitattributes` file

Compared to the default `* text=auto` setting, which results in Windows users checking out text files with CRLF line endings, the `* text eol=lf` setting results in LF line endings on Windows (the desired behavior per `eslintrc.json`).

However, a one-line `.gitattributes` file that only includes `* text eol=lf` tells git to treat all files as text, which can corrupt committed binary files.

Since there's no "auto" mode, it's left to the owner of the git repo to explicitly specify every possible non-text file -- which is why the best practice around line endings is to use a `.gitattributes` file specific for your type of project.

This updated `.gitattributes` file is a slightly modified version of the open-source, community-maintained "Web Project" `.gitattributes` file by @alexkaratarkis from @alexkaratarakis/gitattributes ... Thank you to the contributors.

### Further Reading / References
1. https://gitattributes.io/ and https://github.com/alexkaratarakis/gitattributes
2. https://www.edwardthomson.com/blog/advent_day_1_gitattributes_for_text_files.html
3. https://markoskon.com/line-endings-in-vscode-with-git-and-eslint/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,202 @@
-*		text eol=lf
+## Source: https://github.com/alexkaratarakis/gitattributes
+## Modified * text=auto to * text eol=lf to force LF endings.
+
+## GITATTRIBUTES FOR WEB PROJECTS
+#
+# These settings are for any web project.
+#
+# Details per file setting:
+#   text    These files should be normalized (i.e. convert CRLF to LF).
+#   binary  These files are binary and should be left untouched.
+#
+# Note that binary is a macro for -text -diff.
+######################################################################
+
+# Auto detect
+##   Force LF line endings automatically for files detected as
+##   text and leave all files detected as binary untouched.
+##   This will handle all files NOT defined below.
+*                 text eol=lf
+
+# Source code
+*.bash            text eol=lf
+*.bat             text eol=crlf
+*.cmd             text eol=crlf
+*.coffee          text
+*.css             text
+*.htm             text diff=html
+*.html            text diff=html
+*.inc             text
+*.ini             text
+*.js              text
+*.json            text
+*.jsx             text
+*.less            text
+*.ls              text
+*.map             text -diff
+*.od              text
+*.onlydata        text
+*.php             text diff=php
+*.pl              text
+*.ps1             text eol=crlf
+*.py              text diff=python
+*.rb              text diff=ruby
+*.sass            text
+*.scm             text
+*.scss            text diff=css
+*.sh              text eol=lf
+*.sql             text
+*.styl            text
+*.tag             text
+*.ts              text
+*.tsx             text
+*.xml             text
+*.xhtml           text diff=html
+
+# Docker
+Dockerfile        text
+
+# Documentation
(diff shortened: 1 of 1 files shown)


DoctorDerek:
This is an update to PR #19, which may have caused issue #21.

Hey @timlrx you might want to check if your blog & this template are deploying to Vercel OK, because this was a pretty insidious bug.

My experience was:
1. Nothing happened at all
2. For no clear reason, all the .png files got corrupted
3. Vercel was cool with it, except all the images were broken
4. The images were not broken locally
5. I couldn't diff the images locally, even replacing them with the originals

Once I put in the new `.gitattributes` file, which protects `.png` files among others, and copied back in the original image files, everything works fine.

Hopefully you haven't seen that particular bug (broken images with no warnings -- fun! haha 😅😂), but either way this updated `.gitattributes` file will prevent it from happening.

timlrx:
Ok, thanks for finding out the root cause. I didn't sync the updates to the latest changes so no problems for me :)

DoctorDerek:
Yep, nothing like seeing production look completely different than local
because of something I already know more about than any person really
should ever have to in one lifetime 🤣

Issue #25, opened 2021-04-16
Blog Nested Routes??


mrhut10:
I plan on doing a either a series blog post or dividing a post into bite size chunks so wanted to support nested routing of the blog route.

I've got a local branch that already supports it,

would anyone else benefit from this, should I open a PR??

![image](https://user-images.githubusercontent.com/20330237/115061041-f8b5b480-9f2b-11eb-8941-595e38639a7b.png)

![image](https://user-images.githubusercontent.com/20330237/115061000-e6d41180-9f2b-11eb-8cf0-32558736960a.png)

timlrx:
Thanks @mrhut10 for the new feature! Have merged it to the master branch and released it as v0.3.1 🎉

PR #27, opened 2021-04-16
WIP: Blog sub route support

mrhut10:
relates to #25 

this gives you the ability to sub routing within the blog data.

I wanted to enable so I could do a series of blogs on a particular topic, or break down the blog into smaller sections

image

image


mrhut10:
@timlrx  just discovered a few issues with my above work and now want to make a few improvements.
issues
* migration of slug to an array of strings (paths) feels like bad symantics 
* other areas are references slug property to calculate link targets

Fix Im going to try
* change the page to be defined as `[[path]]/[slug]`
* maybe a new calculated property on posts called `fullPath` which would be either `/blog/[[path]]/slug` or maybe preferably  `/[[path]]/slug`

I won't get time to do so until monday overnight

timlrx:
@mrhut10 , thanks for the feature. I think it would be best to add a section in the readme with the various features for additional use cases e.g. sub-routes, internationalization, multiple authors etc. Would prefer to avoid merging all of it into the template and let people opt in to other features that they would like to have. 

If that sounds good to you, let me know when you are done and I will gladly link your repo. Thanks!

mrhut10:
I didn't even think of all of those use cases, but what a good idea using subroutes for blog translations or sub routes for different authors as well.

yep I'll fix up the improvements I want to already make and document the feature as well when I get back here.
hopefully monday night 🤞

mrhut10:
I've put details into the readme about the nested route,
also the index.xml which i believe is generated for rss purposes, need that be commited? can we remove it and .gitignore it so that its just generated on build?

mrhut10:
I can't see the logs of the above, can you advise why the above preview deploy is failing??

timlrx:
01:23:32.980 | > Build error occurred
-- | --
01:23:32.982 | Error: EISDIR: illegal operation on a directory, read
01:23:32.982 | at Object.readSync (fs.js:568:3)
01:23:32.982 | at tryReadSync (fs.js:353:20)
01:23:32.982 | at Object.readFileSync (fs.js:390:19)
01:23:32.983 | at /vercel/path0/.next/server/pages/tags/[tag].js:758:62
01:23:32.983 | at Array.forEach (<anonymous>)
01:23:32.983 | at getAllTags (/vercel/path0/.next/server/pages/tags/[tag].js:757:9)
01:23:32.983 | at getStaticPaths (/vercel/path0/.next/server/pages/tags/[tag].js:3764:84)
01:23:32.983 | at buildStaticPaths (/vercel/path0/node_modules/next/dist/build/utils.js:16:86)
01:23:32.983 | at /vercel/path0/node_modules/next/dist/build/utils.js:26:628
01:23:32.983 | at processTicksAndRejections (internal/process/task_queues.js:97:5) {
01:23:32.983 | type: 'Error',
01:23:32.983 | errno: -21,
01:23:32.983 | syscall: 'read',
01:23:32.984 | code: 'EISDIR'
01:23:32.984 | }

lib/mdx.js
```diff
@@ -24,10 +24,17 @@ const tokenClassNames = {
   comment: 'text-gray-400 italic',
 }
 
+/**
+ * @param {string} type
```
timlrx, line 28: Can we exclude jsdoc comments? I want to be consistent across the code and don't want to maintain it. Thanks!

timlrx:
The error comes from the line over here - https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/master/lib/tags.js#L14
Probably need to recursively read through the folders.

You can try `npm run build` to build it locally to see if there are any errors.

In general, I think the code current recursive code over the folder is too complicated. You just need to make small edits of the original code. Perhaps define a function in utils like so and use it for the tags and blog page:

```
const getAllFilesRecursively = (dir) => {
  const files = walk(dir)
  const fileStems = files.map(f => f.replace(`${dir}/`, ''))
  return fileStems
}

const walk = (dir) => {
  let files = [];
  fs.readdirSync(dir).forEach(file => {
    let fullPath = path.join(dir, file);
    if (fs.statSync(fullPath).isFile()) {
      files.push(fullPath)
     } else {
      files = files.concat(walk(fullPath));
     }  
  });
  return files
}
```

For the get files part, you can then do

```
const prefixPath = path.join(root, 'data', type)
const files = getAllFilesRecursively(prefixPath)
```

timlrx:
yes, can add index.xml to gitignore

mrhut10:
done the above except had an effort to simplify the getAllFilesRecursively function to still be functional by nows its using functional composition to break down the complexity.

please consider this if thats ok, otherwise can rewrite as per your idea above

timlrx:
All right, thanks. Will just merge it to a subroute branch and make further modifications before merging to master. Cheers!

PR #28, opened 2021-04-16
Update husky to v6

DoctorDerek:
The PR in #23 was incomplete; it was only "working" for me in GitHub Desktop because husky wasn't actually running.

Apparently husky migration from v4 to v6 requires moving the husky scripts from package.json into their own folder: https://typicode.github.io/husky/#/?id=migrate-from-v4-to-v6

I've done so in this PR, and I can confirm that the husky hooks do in fact run from GitHub Desktop with the updated code.

Apparently my specific GitHub Desktop / husky compatibility issue was compounded by something going on with my PATH associated with NVM for Windows, as NVM (for *nix) requires special husky configuration and is thus incompatible with Windows. 😴


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
timlrx, line 1: I don't think this .gitignore is needed?

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
timlrx, line 4: A single line `npx --no-install lint-staged` would do. Use `--no-install` as we have already installed it locally. What's the code above it for?

DoctorDerek:
@timlrx I spent about 6 hours yesterday debugging husky just trying to get it to work, and I don't have the answers to either of those questions...

(Nor do I have an understanding of why npx without --no-install would function differently, though that one at least makes sense to me.)

See the PR I opened @typicode/husky#951 for a discussion of the `.gitignore` file.

The new script is required by husky v6 -- that's how it works now. The docs I linked to in that PR for husky have some more details about it.

timlrx:
Replicated it on my windows and I see that husky automatically adds those lines in to make it work for windows but not for Linux. .gitignore is also created on `husky install`. Ok, that clears things up. Kindly change the lint-staged line to `npx --no-install lint-staged` and I will merge it in.

npx by default installs the package from the npm registry but since we want to ensure that we only use our local lint-staged that we installed we add that flag. Reference: https://www.npmjs.com/package/npx

DoctorDerek:
Solid thanks for the protip @timlrx I appreciate it

PR #31, opened 2021-04-20
add installed but missing eslint-plugin-react-hooks to .eslintrc

wozniaklukasz:
Add installed but missing eslint-plugin-react-hooks to .eslintrc plugins list.


timlrx:
Good find, thanks!

PR #33, opened 2021-04-26
Fix typo

saqibameen:
rename import from `siteMetdata` to `siteMetadata`


DoctorDerek:
Good catch @saqibameen ! I didn't see this one until I converted a fork of this project to TypeScript, and there were so many other changes involved that I forgot to report it.

PR #47, opened 2021-05-17
fix: correct invalid theme toggle on first load

ahmadbilalse:
Bug: Incorrect theme toggle button on first load.
Fix: Added `resolvedTheme` to the conditionals.

On the very first load, the value of `theme` in `ThemeSwitch.js` will be `'system'`. So, the condition `theme === 'dark'` will always be `false` and the site will display the moon button even though it's already in dark mode. In other words, the theme toggle button have to be clicked twice to get to light mode.

You can reproduce the bug by opening the site in incognito.

### DIFF
--- a/components/ThemeSwitch.js
+++ b/components/ThemeSwitch.js
@@ -3,7 +3,7 @@ import { useTheme } from 'next-themes'
 
 const ThemeSwitch = () => {
   const [mounted, setMounted] = useState(false)
-  const { theme, setTheme } = useTheme()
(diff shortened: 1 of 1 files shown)


timlrx:
Good catch, thanks!

stephanlamoureux:
@timlrx This issue seems to be back. After you click through it a couple of times, it corrects itself and seems to stay working if you leave and come back, but it will have this bug for the browser's first visit.

timlrx:
@stephanlamoureux yes, apologies for that. Reverted to using `resolvedTheme` for the fix.

PR #52, opened 2021-05-21
Add pagination support

Music47ell:
I added a new `BlogLayout` so that the `/tags/` page doesn't have pagination.


components/Pagination.js
```diff
@@ -0,0 +1,34 @@
...
+  return (
+    <nav className="flex justify-between">
+      {!prevPage && (
+        <button rel="previous" className="cursor-auto">
+          <strike>Previous</strike>
+        </button>
+      )}
+      {prevPage && (
+        <Link href={currentPage - 1 === 1 ? `/blog/` : `/blog/page/${currentPage - 1}`}>
+          <button rel="previous">Previous</button>
```
timlrx, line 16: How about using the button disabled state instead of strikethrough? Then you can just do 
```
<button rel="previous" disabled={!prevPage}> Previous </button>
```

layouts/BlogLayout.js
```diff
@@ -0,0 +1,96 @@
...
+import Tag from '@/components/Tag'
+import siteMetadata from '@/data/siteMetadata'
+import { useState } from 'react'
+import Pagination from '@/components/Pagination'
+
+const postDateTemplate = { year: 'numeric', month: 'long', day: 'numeric' }
+
+export default function BlogLayout({ posts, title, pagination, prevDisable, nextDisable }) {
+  const [searchValue, setSearchValue] = useState('')
+  const filteredBlogPosts = posts.filter((frontMatter) => {
```
timlrx, line 11: This breaks the search implementation. Ideally, it should work across all the pages. Search results should also be paginated if they exceed a certain length

pages/blog.js
```diff
@@ -1,23 +1,30 @@
 import { getAllFilesFrontMatter } from '@/lib/mdx'
 import siteMetadata from '@/data/siteMetadata'
-import ListLayout from '@/layouts/ListLayout'
+import BlogLayout from '@/layouts/BlogLayout'
 import { PageSeo } from '@/components/SEO'
 
+export const POSTS_PER_PAGE = 3
```
timlrx, line 6: 3 is probably too little? It's good for testing but otherwise 8 or 10 seems more reasonable

pages/blog/page/[page].js
```diff
@@ -0,0 +1,54 @@
+import { PageSeo } from '@/components/SEO'
+import siteMetadata from '@/data/siteMetadata'
+import { getAllFilesFrontMatter } from '@/lib/mdx'
+import BlogLayout from '@/layouts/BlogLayout'
+import { POSTS_PER_PAGE } from '../../blog'
+
+export async function getStaticPaths() {
+  const totalPosts = await getAllFilesFrontMatter('blog')
+  const totalPages = Math.ceil(totalPosts.length / POSTS_PER_PAGE)
+  const paths = Array.from({ length: totalPages - 1 }, (_, i) => ({
```
Comment (timlrx) [inline pages/blog/page/[page].js:10]: `/blog/1` should return the same as `/blog`. Currently, it leads to a 404.

layouts/BlogLayout.js
```diff
@@ -0,0 +1,96 @@
+import Link from '@/components/Link'
+import Tag from '@/components/Tag'
+import siteMetadata from '@/data/siteMetadata'
+import { useState } from 'react'
+import Pagination from '@/components/Pagination'
+
+const postDateTemplate = { year: 'numeric', month: 'long', day: 'numeric' }
+
+export default function BlogLayout({ posts, title, pagination, prevDisable, nextDisable }) {
```
timlrx, line 9: I actually think we do not a new layout and can use the prev one. Just let `pagination` be an optional variable and conditionally render the `Pagination` component. For the tag pages with no pagination passed it, it will display the full list with no pagination (i.e. same behaviour as previously).

timlrx:
The sitemap script also needs to be updated to reflect the new routes. Thanks.

components/Pagination.js
Music47ell: I replaced `strikethrough` with `disabled={!prevPage}` and also added a class with the value of `cursor-auto disabled:opacity-50`

layouts/BlogLayout.js
Music47ell: Right now, the search function searches for a keyword in the current page only.
Is this how you want?
Or, you want the search bar to find keyword and display results from content in `/blog/2` or `/blog/3`?

pages/blog.js
Music47ell: How do you want to approach this? Should users modify the `blog.js` file directly or use the file `siteMetadata.json` to set the amount of posts per pages?

pages/blog/page/[page].js
Comment (Music47ell) [reply pages/blog/page/[page].js]: I achieved this by using the Next.js [Redirect API](https://nextjs.org/docs/api-reference/next.config.js/redirects) and have set the redirect to be `301 Moved Permanently` since it's [better for SEO](https://moz.com/learn/seo/redirection).

layouts/BlogLayout.js
Music47ell: I conditionally rendered the `Pagination` component this using `next/router` and `router.pathname.indexOf('/blog') > -1`

Music47ell:
> The sitemap script also needs to be updated to reflect the new routes. Thanks.

Can you tell me what changes you want for the sitemap script?

timlrx:
> > The sitemap script also needs to be updated to reflect the new routes. Thanks.
> 
> Can you tell me what changes you want for the sitemap script?

The `/blog/page/` routes should be in the sitemap as well

Music47ell:
> > > The sitemap script also needs to be updated to reflect the new routes. Thanks.
> > 
> > 
> > Can you tell me what changes you want for the sitemap script?
> 
> The `/blog/page/` routes should be in the sitemap as well

I no longer use `/blog/page/1`. Now, it's just `/blog/1`.

I also need your feedback [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417981) and [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417946)

layouts/ListLayout.js
```diff
@@ -81,6 +84,17 @@ export default function ListLayout({ posts, title }) {
           })}
         </ul>
       </div>
+      {router.pathname.indexOf('/blog') > -1 && (
```
timlrx, line 87: I still think it is better to check on the pagination prop to decide whether to render this component rather than the router path. This makes it more flexible for someone who wants to also add pagination for tags or not even have pagination for blogs by not passing in the pagination prop

layouts/BlogLayout.js
timlrx: That's because in the prev implementation all the blogs get displayed in a single page. But now that it is split across multiple pages the search for blogs should still cover all blog posts. So the later behaviour would be good. But it should still search only a single page for the tags (because you already filtered on a category)

pages/blog.js
timlrx: Modifying `blog.js` is fine

pages/blog/page/[page].js
Comment (timlrx) [reply pages/blog/page/[page].js]: I am actually thinking of repeating the page with the canonical link set to `/blog`

timlrx:
> > > > The sitemap script also needs to be updated to reflect the new routes. Thanks.
> > > 
> > > 
> > > Can you tell me what changes you want for the sitemap script?
> > 
> > 
> > The `/blog/page/` routes should be in the sitemap as well
> 
> I no longer use `/blog/page/1`. Now, it's just `/blog/1`.
> 
> I also need your feedback [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417981) and [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417946)

Right, in that case the [sitemap](https://tailwind-nextjs-starter-blog-ds56v72u7-timlrx.vercel.app/sitemap.xml) has to include `/blog/1`, `/blog/2` etc.

pages/blog/page/[page].js
Comment (Music47ell) [reply pages/blog/page/[page].js]: This should work now but, because the blog have 8 posts and I set `POSTS_PER_PAGE` to 10, `blog/1` return 404.
Once the blog have more than 10 posts, it will work fine.

timlrx:
There's an issue with the preview deployment. You can click on blog and any one of the post. The error message in the console is: `TypeError: Cannot read property 'currentPage' of undefined`

Music47ell:
> There's an issue with the preview deployment. You can click on blog and any one of the post. The error message in the console is: `TypeError: Cannot read property 'currentPage' of undefined`

Not sure exactly what caused this but the way I solved it was when I moved the `[page].js` file inside a `page` folder.
Do you have any explanation why this solved the issue?

Music47ell:
> There's an issue with the preview deployment. You can click on blog and any one of the post. The error message in the console is: `TypeError: Cannot read property 'currentPage' of undefined`

Not sure exactly what caused this but the way I solved it was when I moved the `[page].js` file inside a `page` folder.
Do you have any explanation why this solved the issue?

Music47ell:
> > > > > The sitemap script also needs to be updated to reflect the new routes. Thanks.
> > > > 
> > > > 
> > > > Can you tell me what changes you want for the sitemap script?
> > > 
> > > 
> > > The `/blog/page/` routes should be in the sitemap as well
> > 
> > 
> > I no longer use `/blog/page/1`. Now, it's just `/blog/1`.
> > I also need your feedback [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417981) and [here](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/52#discussion_r637417946)
> 
> Right, in that case the [sitemap](https://tailwind-nextjs-starter-blog-ds56v72u7-timlrx.vercel.app/sitemap.xml) has to include `/blog/1`, `/blog/2` etc.

Why Do you want to include `/blog/page/1`, `/blog/page/2` in the sitemap?

I've checked a sitemap created using Yoast SEO plugin on a WordPress site with pagination and pagination pages aren't included in the sitemap.

And right now, when you generate a sitemap using your script, it has the same behaviour.

DoctorDerek:
Solid thanks for checking ☺️

Best regards,
Derek

timlrx:
Thanks, learnt something new about pagination and sitemaps! No need for the change in that case

Music47ell:
@DoctorDerek @timlrx You are both welcome.

timlrx:
2 small things left:
1) Could you change the `ListLayout` to conditionally render the pagination based on the pagination prop rather than the route, as previously discussed
2) resolve the merge conflict on package-lock

If you could restore the search functionality to work on all the blog posts that would be great, otherwise I will patch back that functionality in a subsequent PR. The tricky part is probably handling search results that exceed the max page per post size. I think we should just let it exceed (i.e. not paginate) like the tags page and also not display the pagination component.

Music47ell:
> 2 small things left:
> 1) Could you change the `ListLayout` to conditionally render the pagination based on the pagination prop rather than the route, as previously discussed
> 2) resolve the merge conflict on package-lock
> 
> If you could restore the search functionality to work on all the blog posts that would be great, otherwise I will patch back that functionality in a subsequent PR. The tricky part is probably handling search results that exceed the max page per post size. I think we should just let it exceed (i.e. not paginate) like the tags page and also not display the pagination component. 

I just finished working on the `ListLayout` locally and I'll push a commit a little bit later. 

I will also resolve the merge conflict. 

As for search, I might have a solution for this. I'll work on it as soon as possible. I'll let you know if it works or not.

Music47ell:
Everything should be ready now. Pagination component will render conditionally, and `package-lock` conflict is resolved.

Right now, pagination will work only with the blog. I didn't implement it in the `[tag].js` page.

timlrx:
Thanks, LGTM! Will merge it to `feat/pagination` first and once the search issue is resolved, we can merge to master.

PR #69, opened 2021-06-12
Add convertStringToHTML function to prevent RSS encoding issues

DLevai94:
When parsing descriptions and titles with characters like '&', '<', etc. feed was broken. This regex swaps these characters with their HTML entity equivalent.

### DIFF
--- a/lib/generate-rss.js
+++ b/lib/generate-rss.js
@@ -1,11 +1,14 @@
 import siteMetadata from '@/data/siteMetadata'
 
+const convertStringToHTML = (string) =>
+  string.replace(/&/g, '&amp;').replace(/>/g, '&gt;').replace(/</g, '&lt;').replace(/"/g, '&quot;')
+
 const generateRssItem = (post) => `
   <item>
     <guid>${siteMetadata.siteUrl}/blog/${post.slug}</guid>
-    <title>${post.title}</title>
+    <title>${convertStringToHTML(post.title)}</title>
     <link>${siteMetadata.siteUrl}/blog/${post.slug}</link>
-    <description>${post.summary}</description>
+    <description>${convertStringToHTML(post.summary)}</description>
     <pubDate>${new Date(post.date).toUTCString()}</pubDate>
     <author>${siteMetadata.email} (${siteMetadata.author})</author>
     ${post.tags.map((t) => `<category>${t}</category>`).join('')}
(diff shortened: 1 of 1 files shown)


DoctorDerek:
Thanks @DLevai94 !!

Looks good to me, though i could imagine certain other characters not being RSS safe. I'm not sure what those would be, per se, and this would seem an improvement!

timlrx:
Thanks, to @DoctorDerek point of other unsafe characters, how about using [encodeURIComponent()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)? To deal with the unescaped characters we can just replace them using this snippet I found on [stackoverflow](https://stackoverflow.com/questions/75980/when-are-you-supposed-to-use-escape-instead-of-encodeuri-encodeuricomponent):

```
function encodeData(s:String):String{
    return encodeURIComponent(s).replace(/\-/g, "%2D").replace(/\_/g, "%5F").replace(/\./g, "%2E").replace(/\!/g, "%21").replace(/\~/g, "%7E").replace(/\*/g, "%2A").replace(/\'/g, "%27").replace(/\(/g, "%28").replace(/\)/g, "%29");
}
```

If you can verify that this works for you and the RSS readers, I would be happy to merge it in.

DLevai94:
> If you can verify that this works for you and the RSS readers, I would be happy to merge it in.

I think encodeURIComponent is not a great use-case for this, because we don't have problems with most of the characters, like _ or a space, but the special characters from XML only, like <, >, ", &.

But I've checked it out just in case, this creates a feed like this:

```
<item>
    <guid>https://davidlevai.com/tailwind-glassmorphism</guid>
    <title>Glassmorphism%20with%20Tailwind%20CSS%20Under%2060%20seconds</title>
    <link>https://davidlevai.com/tailwind-glassmorphism</link>
    <description>Quick%20guide%20on%20how%20to%20achive%20the%20hyped%20glass%20morphism%20effect%20with%20Tailwind%20%28updated%20with%20Tailwind%20v2%2E1%29</description>
    <pubDate>Thu, 17 Dec 2020 00:00:00 GMT</pubDate>
    <author>david@davidlevai.com (David Levai)</author>
    <category>react</category><category>tailwind</category><category>coding</category>
  </item>
```

DoctorDerek:
Great point and thanks for testing Dávid.

According to Señor Google there are 5 XML special characters. I thought at
first / might be one, but obviously not.
https://docs.oracle.com/cd/A97335_02/apps.102/bc4j/developing_bc_projects/obcCustomXml.htm

So it would seem the original PR code would be robust.

timlrx:
Sounds good, thanks for verifying!

PR #71, opened 2021-06-18
Remove Unexpected File

Teng:
If there is an unexpected file error, eg `.DS_Store` common in macOS

I'm a beginner, but this works 😊

```
TypeError: Cannot read property 'join' of undefined
    at eval (webpack-internal:///./layouts/ListLayout.js:37:86)
    at Array.filter (<anonymous>)
    at ListLayout (webpack-internal:///./layouts/ListLayout.js:36:35)
```
```
TypeError: Cannot read property 'map' of undefined
    at eval (webpack-internal:///./pages/index.js:134:40)
    at Array.map (<anonymous>)
    at Home (webpack-internal:///./pages/index.js:72:84)
```
image

### DIFF
--- a/lib/mdx.js
+++ b/lib/mdx.js
@@ -98,6 +98,10 @@ export async function getAllFilesFrontMatter(folder) {
   files.forEach((file) => {
     // Replace is needed to work on Windows
     const fileName = file.slice(prefixPaths.length + 1).replace(/\\/g, '/')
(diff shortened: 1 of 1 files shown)


timlrx:
Thanks! I remember seeing many issues on this in the past with mac users so hopefully, this is the last :)

PR #73, opened 2021-06-18
Swap convertStringToHTML with html-escaper module

Music47ell:
I replaced the new `convertStringToHTML` function with the [html-escaper](https://www.npmjs.com/package/html-escaper) module [(GitHub repo)](https://github.com/WebReflection/html-escaper).

According to the developer of this module:
> do not ever replace one char after another if you are transforming a string into another

Because it might produce unexpected/unwanted results [like this](https://github.com/WebReflection/html-escaper#why).

### DIFF
--- a/lib/generate-rss.js
+++ b/lib/generate-rss.js
@@ -1,4 +1,4 @@
-import { escape } from 'html-escaper'
+import { escape } from '@/lib/utils/htmlEscaper'
 
 import siteMetadata from '@/data/siteMetadata'
 
--- a/lib/generate-rss.js
+++ b/lib/generate-rss.js
@@ -1,14 +1,13 @@
-import siteMetadata from '@/data/siteMetadata'
+import { escape } from 'html-escaper'
 
-const convertStringToHTML = (string) =>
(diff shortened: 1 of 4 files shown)


DoctorDerek:
Sounds good, nice catch. I always end up with bizarre results when I'm
working with emojis in strings, so it'd make sense there would be a bug in
just "straight up" character replacement.

timlrx:
Thanks! Can we just copy over the escape function to `lib/utils`? Would prefer not to have an additional dependency for a small utility function.

DoctorDerek:
Isn't the point of the npm ecosystem to add dependencies that can be
maintained instead of making one-off utils? I'm just curious what the
advantage would be.

timlrx:
That's the story of left-pad...

Music47ell:
> That's the story of left-pad...

Oh yeah, I remember this.
For those who are interested in reading about left-pad more, you can read the [Arstechnica](https://arstechnica.com/information-technology/2016/03/rage-quit-coder-unpublished-17-lines-of-javascript-and-broke-the-internet/) article and [npm blog post](https://blog.npmjs.org/post/141577284765/kik-left-pad-and-npm.html).

I just imported the `escape` function locally and tested it.

timlrx:
LGTM, thanks!

PR #79, opened 2021-06-19
Compose.js Updates

AryaveerSR:
Added a more interactive Compose.js Prompt using Inquirer.js and handling for empty values


AryaveerSR:
forgot updaing readme, rip

PR #93, opened 2021-06-24
Add JSON-LD with Article structured data support

Music47ell:


Music47ell:
Google Rich Results Test return 0 warnings and errors.

Test results link: https://search.google.com/test/rich-results?id=RZt5D53-O5LsdE2Qu0rOIw

components/SEO.js
```diff
@@ -37,11 +37,37 @@ export const BlogSeo = ({ title, summary, date, lastmod, images = [] }) => {
...
+      '@type': 'WebPage',
+      '@id': url,
+    },
+    headline: title,
+    image: featuredImages,
+    datePublished: publishedAt,
+    dateModified: modifiedAt,
+    author: {
+      '@type': 'Person',
+      name: siteMetadata.author,
```
timlrx, line 58: `author` should be a list of authors since now we support multiple authors and taken from the frontmatter. If none is provided we can fall back to `siteMetadata.author`.  To verify that it displays correctly you can try it on this page - https://tailwind-nextjs-starter-blog-7uibpj70o-timlrx.vercel.app/blog/introducing-tailwind-nextjs-starter-blog

components/SEO.js
```diff
@@ -37,11 +37,42 @@ export const BlogSeo = ({ title, summary, date, lastmod, images = [] }) => {
...
     }
   })
 
+  let authorsArr = authors.length === 0 ? [siteMetadata.author] : authors
+  const authorList = authorsArr.map((author) => {
+    return {
+      '@type': 'Person',
+      name: author,
+    }
+  })
```
Music47ell, line 51: I did it this way but now JSON-LD pull the name from the frontMatter of the blog post and not the actual name from the authors Markdown files.

Example: For the blog post with the multi-authors, JSON-LD now look like this:

```json
"author": [
    {
      "@type": "Person",
      "name": "default"
    },
    {
      "@type": "Person",
      "name": "sparrowhawk"
    }
  ]
```

As you can see, the results show `'default'` and `'sparrowhawk'` instead of `'Tails Azimuth'` and `'Sparrow Hawk'`.

timlrx: Should have pointed you to the [authorDetails prop](https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/v1/layouts/PostLayout.js#L17) which gets fed into every blog page. Just need to add it in the blogSeo component and the names should be correct.

Music47ell: I knew about `authorDetails` and I tried it multiple times yesterday and it didn't work. I kept getting the error message: `TypeError: Cannot read property 'map' of undefined`.

The issue was that I forgot to add `authorDetails={authorDetails}` on [this line](https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/e92b7d2737aae0d9bec2941880f6f96f3ff869b1/layouts/PostLayout.js#L24)

Now, everything should work fine.

timlrx:
Looks good, thanks for the help!

timlrx (approved):

PR #96, opened 2021-06-29
feat: mdx-bundler and xdm

timlrx:
This PR switches to the new xdm and mdx-bundler MDX compiler and bundler libraries. It updates all remark, rehype packages to micromark 3 compatible ones. It also introduces a new rehype-prism-plus library which supports line highlighting and line numbers, and a custom implementation of remark-code-title to work with the newest version of remark.


components/Card.js
```diff
@@ -1,5 +1,5 @@
-import Image from 'next/image'
-import Link from '@/components/Link'
+import Image from './Image'
+import Link from './Link'
```
Music47ell, line 2: Is it better to import the `Link` component this way instead of `@/components/Link`?

timlrx: It's just my preference to import with the relative path if they are in the same folder

PR #99, opened 2021-06-30
Migrate v1 to typescript

frontendwizard:
I've seen the discussion on #66 about typescript and I thought it'd be fun to try so here it is. I'm not sure I got everything as they should be, mdx.js typing was specially tricky but, this is as far as I went. If y'all feel like, I'd love to hear on how could this be improved and maybe even merged into the repo, but feel free to close it if it's really not somewhere you'd like to go right now on v1.


frontendwizard:
There we go. Everything is working now. I tried to avoid type assertion as much as I could've, but there are some points where I couldn't find a way out, like when typing the result of the `matter` package. I'm open to suggestions if anyone has any.

timlrx:
Thanks for the contribution! This is really great, but as mentioned in the discussion, I still prefer to keep this in js and maintain one code base. I would happily add a link to your repository in the readme so that potential users can find a ts version of the code if they are interested in that. Let me know if that is good for you?

frontendwizard:
Works for me, I don't mind maintaining a fork of this migrated to the typescript version. When y'all decide moving to it, I can reopen a pull request from it.

frontendwizard:
my only concern at this moment is how to deal with `v1`/`master` and other branches. Should I move the typescript setup to the `master` branch and then merge it into each separate branch? Should I just hang on the `v1` branch until it gets merged into `master`? Do you have any estimations on when is `v1` going to happen?

timlrx:
I plan on merging `v1` to `master` either this week or next. The only thing left pending is finalizing the changes on `mdx-bundle` and better documentation.

Rather than maintain a fork, how about maintaining a separate branch? We can call the branch `typescript`  and I can add you as a maintainer if you are on board with it.

frontendwizard:
That makes sense, I believe it's more effective. I can do that.

frontendwizard:
I'll update the setup instructions to something like
```
npx degit timlrx/tailwind-nextjs-starter-blog#typescript
```

frontendwizard:
once you have setup the branch I can point this PR to it :)

timlrx:
v1 is ready! Added you as a collaborator as well so you should be able to directly create and edit of a new branch, thanks. I made some changes to the js template so once you resolve those, we should be good.

aloisdg (approved):

frontendwizard:
Everything seems to be good, except for the about page. Whenever I try to access it, I get a mismatch version log from esbuild on the console.
```
event - compiled successfully
 > error: Cannot start service: Host version "0.12.14" does not match binary version "0.11.23"
```

timlrx:
Try deleting your `node_modules` and `package-lock.json` and re-installing the dependencies. This is caused by conflicting esbuild versions used in the existing repo and the v1 branch.

frontendwizard:
Types can be improved still, but it's working as is now so I'm merging in.

PR #141, opened 2021-07-21
refactor: use env variables for comments system

Music47ell:


timlrx (approved):
LGTM, thanks! Have added the env variable in vercel as well so let's merge it and it should work on master.

PR #155, opened 2021-07-29
fix: rename `giscus-frame` class to `giscus`

### LINKED ISSUE #154: Comment Box is not showing properly in Window 10
**To Reproduce**
1. Go to 'https://tailwind-nextjs-starter-blog.vercel.app/blog/nested-route/introducing-multi-part-posts-with-nested-routing' and scroll down to the comment section.
2. Click on 'Load comments.'
3. See error

**Screenshots**
image

**System Info
 - Window 10

**Browser Info (if display / formatting issue):**
- Device: Desktop
- Browser chrome
- Version: Version 91.0.4472.164 (Official Build) (64-bit)

laymonage:
Fix #154.

The error:

image

It is caused by a recent update in giscus: https://github.com/laymonage/giscus/commit/ae401f85cea100381554f7a739cb7fe7a415f2b4

The update was done to prevent `iFrameResizer` from interfering with other `<iframe>` elements on the page. (I think by default it applies to all `<iframes>` that exist on the page.)

However, the error exists because you use the `giscus-frame` class for the div. The `giscus-frame` is meant to be used by giscus' `<iframe>` tag only, and it's managed by the client script. You should use the `giscus` class instead to tell giscus where to put the `<iframe>`.

From the docs:

image

### DIFF
--- a/components/comments/Giscus.js
+++ b/components/comments/Giscus.js
@@ -42,7 +42,7 @@ const Giscus = ({ mapping }) => {
   return (
     <div className="pt-6 pb-6 text-center text-gray-700 dark:text-gray-300">
       {enableLoadComments && <button onClick={LoadComments}>Load Comments</button>}
-      <div className="giscus-frame" id={COMMENTS_ID} />
+      <div className="giscus" id={COMMENTS_ID} />
     </div>
   )
 }


timlrx:
Thanks for the quick fix!

PR #164, opened 2021-08-05
chore(slug): added ssr for other locales

GautierArcin:
Hello,

I found a little something on you pr for the translation demo. 

The `locale` should be outside  `params` scope, Next js documentation is very unclear on that point. That's why you had to enable the Fallback.

(Thank you a lot for the demo ! It's going to be very useful for my own blog !)


timlrx:
Thanks! That solves the remaining piece that I was puzzling about last night

PR #165, opened 2021-08-06
A locales menu and a localization switch have been added

rsipakov:
A locales menu and a localization switch have been added for a clear demonstration translation.


timlrx:
Do you mind merging your demo with next-translate instead? And maybe without the `.idea` files as well. Alternatively, I would be happy to link to your repo instead as a complete example in the readme.

rsipakov:
p.s. Replaced with another pull request with an integrated package "next-translate".

PR #181, opened 2021-08-09
fix: toc issue with remark-autolink-headings

Music47ell:
TOC won't work if `remark-autolink-headings` option `behavior` is set to `append`


timlrx:
Thanks, makes sense to me. You are running a version with `remark-autolink-headings` set to `append`?

Music47ell:
You're welcome. Yes, and I couldn't get the TOC to work until I made those changes.

PR #182, opened 2021-08-09
feat: add topic rss feed to tag seo

GautierArcin:
Rss feed on tags were not working (it seems).

 I created another component in SEO, the only difference being that they add a link to the rss tags.

### DIFF
--- a/components/SEO.js
+++ b/components/SEO.js
@@ -24,6 +24,34 @@ export const PageSeo = ({ title, description }) => {
   )
 }
 
+export const TagSeo = ({ title, description }) => {
+  const router = useRouter()
+  return (
+    <Head>
+      <title>{`${title}`}</title>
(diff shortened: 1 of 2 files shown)


timlrx:
I guess you meant to suggest including the RSS feed for that particular topic as an alternate feed in the tag page? Previously, I did not include it as I used it just for submitting it to curated RSS feed readers but I guess we can include it.

GautierArcin:
Yup ! I was suggesting indeed to include the RSS feed for that particular topic as an alternate feed. You can propose multiple feed for each page.

The main feed was working. It's just that RSS tag feed were, (I think but I might be wrong), `inaccessible` (i.e. you could access them with direct url but there were not any button or SEO that linked them. You "" had to know "" they where there).

PR #187, opened 2021-08-10
refactor: reduce duplication and make seo title consistent

Music47ell:


timlrx:
Hi @Music47ell, any reason for the refactor? `taxonomy` is not the clearest term and I would prefer is people do not turn it on for random pages where there might not be a valid rss feed. Hence, we scoped it only to the Tag page. If you are just trying to cut down on repetition, then how about creating a `CommonSEO` component and using it across `PageSEO`, `BlogSEO` and `TagSEO`?

Music47ell:
Hi @timlrx, I am trying to cut down on repetition. As for the use of the word `taxonomy`, it is the term used by WordPress and Hugo for Tags and Categories, which is what I use on my site.

Since this template only uses Tags, I just renamed it back to TagSEO.

Do you want me to move the meta tags in the `_document.js` file as well to the new `CommonSEO` component?

timlrx:
Thanks for the clarification. In that case, could you help to address #188 as well? For consistency, let's change the title in `TagSEO` to use `${siteMetadata.author}` instead of `${siteMetadata.title}`. Contrary to my suggestion over there, we can just keep `[...slug]` as it is and use only the post slug title.

No need to move the meta tags from `_document.js` as they should apply to all pages even for ones where a user does not want to include the SEO header.

Music47ell:
You're welcome. I'll implement the changes you requested and commit them soon.

timlrx:
LGTM, thanks! @GautierArcin are you good with the changes as well? If so, I will merge it in.

GautierArcin:
@timlrx  That's good for me ! Thanks a lot @Music47ell for this change !

pages/projects.js
```diff
@@ -1,14 +1,12 @@
...
-import Link from '@/components/Link'
 import Card from '@/components/Card'
-import { PageSeo } from '@/components/SEO'
+import { PageSEO } from '@/components/SEO'
 
 export default function Projects() {
   return (
     <>
-      <PageSeo title={`Projects - ${siteMetadata.author}`} description={siteMetadata.description} />
+      <PageSEO title={`Projects - ${siteMetadata.author}`} description={siteMetadata.description} />
```
GautierArcin, line 9: Sorry, I just dig a bit deeper because I updated my i18n repo.

What would you think if we replaced `description={siteMetadata.description}` with `description={"Things I created"}` ? It's very very minor, but I think it could be more coherent with `tags`. `siteMetadata.description` would be left only on pages where we can access blogs posts.

timlrx: I think it is fine as it is. I changed the description of the tags page to "Things I blog about" just to add a bit of flare band hopefully nudge the user to want to modify some of the SEO tags himself.

PR #198, opened 2021-08-20
fix: non-latin tags don't cause error

### LINKED ISSUE #197: Unable to have cyrillic tags
**Describe the bug**
`kebabCase` function doesn't work with non-Latin characters.

**To Reproduce**
1. Go to any blog post
2. Add any tag with non-Latin characters, for example `производительность` (performance in Russian)
3. See error

**Expected behavior**
Build passed and tag correctly works

**Screenshots**
CleanShot 2021-08-18 at 20 21 42@2x

**System Info (if dev / build issue):**
 - OS: macOS
 - Node version: 14 LTS
 - Npm version: -

vitalybaev:
This MR force to use `github-slugger` for tags. This allows non-latin tags to work properly.

Closes #197

### DIFF
--- a/lib/utils/kebabCase.js
+++ b/lib/utils/kebabCase.js
@@ -1,8 +1,5 @@
-const kebabCase = (str) =>
-  str &&
-  str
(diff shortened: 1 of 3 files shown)


timlrx:
Thanks for the fix!

Issue #209, opened 2021-08-29
Code Block: no spaces on the right side


renomureza:
**Describe the bug**

There are no spaces on the right side of the code block.

**Expected behavior**

The space on the right side must be the same as the left side.

**Screenshots** https://tailwind-nextjs-starter-blog.vercel.app/blog/new-features-in-v1

![image](https://user-images.githubusercontent.com/49445216/131235887-49351536-a6f0-4a55-9488-45efbf159e86.png)

timlrx:
Agree that it would look better with some paddiwng on the right. Would gladly accept a PR on this if someone wants to contribute a fix!

larbisahli:
@timlrx A small hack.

```css
/* globals.css */
.code-line > span:last-child {
    padding-right: 4px;
    }
```

timlrx:
Fix in #216. Thanks @llovvoll for trying it out and @larbisahli for finding a fix that should work for all cases.

renomureza:
This seems to be causing the line highlighting not to fully highlight the line.

Before: https://tailwind-nextjs-starter-blog-git-fork-humble-barn-a10bc2-timlrx.vercel.app/blog/new-features-in-v1#line-highlighting-and-line-numbers

![image](https://user-images.githubusercontent.com/49445216/132113562-c833ba11-5e83-439d-b0ca-523fd075617e.png)

After: https://tailwind-nextjs-starter-blog.vercel.app/blog/new-features-in-v1#line-highlighting-and-line-numbers

![image](https://user-images.githubusercontent.com/49445216/132113540-353ad46b-b90c-4782-801c-99a15f033b75.png)

timlrx:
Yes, you are right, `w-max` breaks line highlighting.

timlrx:
I will modify `rehype-prism-plus` plugin directly to wrap text nodes in a span so that the solution proposed by @llovvoll would work for all cases:

```
.code-line > span:last-child {
  @apply pr-4;
}
```

renomureza:
Yes it seems to be the best solution and it can solve the same problem on mobile.

If possible we need to change the `div` with `span` to make it valid validator.w3.org like Tailwindcss documentation:: https://validator.w3.org/nu/?doc=https%3A%2F%2Ftailwindcss.com%2Fdocs%2Ffloat

Without error: _Error: Element div not allowed as child of element code in this context._

[...]

PR #212, opened 2021-08-29
fix: code block space #209

llovvoll:
Hi, 

I'm not sure if the pr solves #209 issue

### DIFF
--- a/css/tailwind.css
+++ b/css/tailwind.css
@@ -18,7 +18,7 @@
   @apply pl-4 -mx-4 border-l-4 border-gray-800;
 }
 
-.code-line > .text-code-white:last-child {
+.code-line > span:last-child {
   @apply pr-4;
 }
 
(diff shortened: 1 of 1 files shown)


timlrx:
Works for the example in the issue but does not work for this one:
image

Instead of targeting `text-code-white` try targeting the last `span`?

llovvoll:
> Works for the example in the issue but does not work for this one:
> image
> 
> Instead of targeting `text-code-white` try targeting the last `span`?

Sorry, I missed that point, I have updated the commit

timlrx:
image

I think we have to handle the case where the last element is not wrapped in a span:

```
<div class="code-line">
<span class="text-code-blue">git</span>
 remote 
<span class="text-code-blue">add</span>
 template git@github.com:timlrx/tailwind-nextjs-starter-blog.git
</div>
```

llovvoll:
Agree, It may take time to read remarkJS document

renomureza:
Not all have span, example: https://tailwind-nextjs-starter-blog.vercel.app/blog/new-features-in-v1#configuring-a-blog-post-frontmatter

llovvoll:
I think this pr can close first.

PR #216, opened 2021-08-30
fix: code block space #209

larbisahli:
Clean fix for the spaces on the right side of the code block.


nguyenyou (approved):

timlrx (approved):
This does the trick, thanks!

PR #227, opened 2021-09-05
Fix code highlighting issues and update linter

### LINKED ISSUE #209: Code Block: no spaces on the right side

timlrx:
- Fix #209 
- Fix highlighted code block that does not span the full width when overflow
- Use `span` instead of `div` for code line
- Update linter and fix lint issues


components/analytics/GoogleAnalytics.js
```diff
@@ -10,7 +10,7 @@ const GAScript = () => {
         src={`https://www.googletagmanager.com/gtag/js?id=${siteMetadata.analytics.googleAnalyticsId}`}
       />
 
-      <Script strategy="lazyOnload">
+      <Script strategy="lazyOnload" id="ga-script">
```
nguyenyou, line 13: Do we need to provide the `id` for the other `<Script />` as well?

timlrx: It's recommended for inline scripts.

> Inline scripts wrapped with <Script> require an id attribute to be defined to track and optimize the script.

The ones which I modified are those picked up by the linter.

PR #267, opened 2021-11-06
fix: make line numbers in code block right aligned for a consistent UI

PsyGik:
## Before

When enabling line numbers using `showLineNumbers` in a MD code block, the line numbers were left aligned causing the code from Line 1-9 to be slightly on the left than for code from L10 onwards. 

image

## After
This PR fixes the above bug and right aligns the line numbers (following IDE conventions), and if merged code blocks will look like 

image

--- 
## Reasoning

This also follows the same way IDEs represent line numbers,

- VSCode:
image

- WebStorm:
image

### DIFF
--- a/css/prism.css
+++ b/css/prism.css
@@ -28,7 +28,7 @@
 }
 
 .line-number::before {
-  @apply pr-4 -ml-2 text-gray-400;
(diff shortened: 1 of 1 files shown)


timlrx:
Thanks!

PR #287, opened 2021-12-08
Improve starter experience

huydhoang:
Hi,
I've updated the `npx degit` and `npm start` code blocks to have separate lines.
With that, starters can copy a command without manually highlighting it with their mouse.
I hope it makes sense.


timlrx (approved):

timlrx:
Makes sense to me, thanks

Issue #289, opened 2021-12-12
Disable dark theme


gutsytechster:
I'd like to disable the dark theme temporarily. I hid the toggle button to switch the theme. However, the mobile view by default shows the dark theme, where can I configure that?

PS: Thanks for the project. It was an easy-to-use & configurable template.

timlrx:
Thanks, it uses `system` by default so probably your mobile phone is configured to prefer dark theme. In any case, @llovvoll added a new `theme` setting in the config file which you can change to `light` to force it to prefer the light theme instead. You might have to clear your cookies to see the changes.

gutsytechster:
Thank you for the work! I'll test it out, and will close this issue if everything works fine.

timlrx:
All's good? If so, I will close the issue.

PR #291, opened 2021-12-15
feat: customize default theme

llovvoll:
Hi,

Now you can choose the default theme in siteMetadata

#289


timlrx:
Will change the default theme to 'system' in the main template, but thanks for the contribution!

PR #301, opened 2021-12-19
Use '::' for pseudo-classes & ::marker

AlexanderZeilmann:
This pull request encompasses three closely related changes.
See the changes live here https://tailwind-nextjs-starter-blog-alexanderzeilmann.vercel.app/blog/github-markdown-guide

## `::before` instead of `:before` 
In CSS both `::before` and `:before` work, but `::before` is preferred.
Tailwind, however, accepts only the double-colon syntax, i.e. `'code:before': {content: 'none', }`
has no effect at all in `tailwind.config.js`, but `'code::before': {content: 'none', }` does.

**Changes: Replace `:before` and `:after` by `::before` and `::after`**

Rendering before the change:
colon_before
Rendering after the change:
colon_after

## `::marker` additional to `::before` in lists in CSS
Using the `before`-pseudo element to style the list-items seems to be not working in current browsers.
The new way is using the `::marker` pseudo element.

**Changes: Add the following in `tailwind.css`**

```css
.task-list-item {
  @apply list-none;
}
```

Rendering before the change (note the gray bullet-points):
before

Rendering after the change:
after

## `::marker` in Tailwind
[Tailwind uses the `::marker` pseudo-element for styling lists.](https://github.com/tailwindlabs/tailwindcss-typography/blob/b3e8c22d11d549620c3b5f27e883726929a57a51/src/styles.js#L1305-L1311)

Using 
```js
'ol li::before': {
  fontWeight: '600',
  color: theme('colors.gray.500'),
},
```
in `tailwind.config.js` does not overwrite the `::marker` use by tailwind. For this we need:
```js
'ol li::marker': {
  fontWeight: '600',
  color: theme('colors.gray.500'),
},
```

**Changes: replace `ol li:before` and `ul li:before` by `ol li::marker` and `ul li::marker` in `tailwind.config.js`**

### DIFF
--- a/css/tailwind.css
+++ b/css/tailwind.css
@@ -2,8 +2,12 @@
 @tailwind components;
 @tailwind utilities;
 
-.task-list-item:before {
+.task-list-item::before {
   @apply hidden;
+}
+
+.task-list-item {
+  @apply list-none;
 }
 
 html {
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -77,16 +77,9 @@ module.exports = {
               borderRadius: '0.25rem',
             },
             hr: { borderColor: theme('colors.gray.200') },
-            'ol li::before': {
-              fontWeight: '600',
-              color: theme('colors.gray.500'),
-            },
             'ol li::marker': {
               fontWeight: '600',
               color: theme('colors.gray.500'),
-            },
-            'ul li::before': {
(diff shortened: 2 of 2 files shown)


timlrx:
Thanks for spotting this! This must be due to the changes in `tailwindcss-typography` introduced during the upgrade to tailwind v3.

PR #307, opened 2021-12-22
Add security headers - fixes #304

AlexanderZeilmann:
For details see #304

Issue #304, opened 2021-12-20
Add security HTTP response headers


AlexanderZeilmann:
**Is your feature request related to a problem? Please describe.**
[The `tailwind-nextjs-starter-blog` template does not define any security headers currently.](https://securityheaders.com/?q=https%3A%2F%2Ftailwind-nextjs-starter-blog.vercel.app%2F&hide=on&followRedirects=on)

**Describe the solution you'd like**
[Next.js makes it possible to define custom security headers.](https://nextjs.org/docs/advanced-features/security-headers)

The `X-Frame-Options`, `X-Content-Type-Options` and the `Referrer-Policy` headers should be always possible. 
The `Content-Security-Policy` and the `Permissions-Policy` header would also be nice, but it might be necessary for users to change them as they develop their website.
The `Strict-Transport-Security` header is automatically added by Vercel, but other hosters might not include it.

**Describe alternatives you've considered**
For some of the headers there are no alternatives. And even if there were alternatives, including them would be a good idea for a better security.

timlrx:
Yes, this would be a nice addition. I modified a baseline off [Lee Robinson's blog](https://github.com/leerob/leerob.io/blob/af1fc3f3ce168fa351c46076a7a654d97db8f016/next.config.js)

As you mentioned, there would be some hassle involved if someone were to add external services, so we would probably have to add that to the readme, but otherwise I think this is a good start?

The alternative is to add it as a [middleware](https://github.com/leerob/leerob.io/blob/main/pages/_middleware.ts). Have not tried the middleware approach. Do you know whether it applies to static built sites as well? If it does not, then I think the preference would be to modify the config.

```
// You might need to insert additional domains in script-src if you are using external services
const ContentSecurityPolicy = `
  default-src 'self';
  script-src 'self' 'unsafe-eval' 'unsafe-inline';
  style-src 'self' 'unsafe-inline' *.googleapis.com;
  img-src * blob: data:;
  media-src 'none';
  connect-src *;
  font-src 'self' 'fonts.gstatic.com';
`;

const securityHeaders = [
  // https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
  {
    key: 'Content-Security-Policy',
    value: ContentSecurityPolicy.replace(/\n/g, '')
  },
  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
  {
    key: 'Referrer-Policy',
[...]

PR #308, opened 2021-12-23
fix: can not load external script & Unable to get disqus config of shorname

llovvoll:
Screen Shot 2021-12-24 at 1 08 35 AM

1. Fix unable to load external js (Google Analytics and comment function), because Content Security Policy did not add the above to the whitelist #307 
Reproduce: https://tailwind-nextjs-starter-blog.vercel.app/blog/new-features-in-v1

2. Fix unable to get disqus config of shorname

### DIFF
--- a/components/comments/Disqus.js
+++ b/components/comments/Disqus.js
@@ -16,7 +16,7 @@ const Disqus = ({ frontMatter }) => {
     }
     if (window.DISQUS === undefined) {
       const script = document.createElement('script')
-      script.src = 'https://' + siteMetadata.comment.disqus.shortname + '.disqus.com/embed.js'
+      script.src = 'https://' + siteMetadata.comment.disqusConfig.shortname + '.disqus.com/embed.js'
       script.setAttribute('data-timestamp', +new Date())
       script.setAttribute('crossorigin', 'anonymous')
       script.async = true
--- a/next.config.js
+++ b/next.config.js
@@ -5,12 +5,13 @@ const withBundleAnalyzer = require('@next/bundle-analyzer')({
 // You might need to insert additional domains in script-src if you are using external services
 const ContentSecurityPolicy = `
   default-src 'self';
-  script-src 'self' 'unsafe-eval' 'unsafe-inline';
+  script-src 'self' 'unsafe-eval' 'unsafe-inline' giscus.app;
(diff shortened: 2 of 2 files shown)


AlexanderZeilmann:
Ahh right, I forgot them, sorry 🙈 
While we are at it, maybe we should also add the URLs for the other analytics scripts (`plausible.io`, `scripts.simpleanalyticscdn.com` and `vitals.vercel-insights.com`).

llovvoll:
> Ahh right, I forgot them, sorry 🙈 While we are at it, maybe we should also add the URLs for the other analytics scripts (`plausible.io`, `scripts.simpleanalyticscdn.com` and `vitals.vercel-insights.com`).

I have updated the whitelist, thank you for your contribution 🍻

AlexanderZeilmann:
I just figured out that we do not have to put `vitals.vercel-insights.com` in the CSP, as we do not load scripts from there but only send pings, which is already allowed by `connect-src *;`.
I thought that the CSP was blocking `vitals.vercel-insights.com` on my site, but it was actually my ad-blocker 🙄

timlrx:
I preference would be to only add `giscus.app` to the CSP so that it works for this demo. Users who clone the template would have to configure the CSP based on other integrations which they opted in for. Yes, it's a little more hassle this way but I think it's better than whitelisting every analytics or comment provider.

llovvoll:
> I preference would be to only add `giscus.app` to the CSP so that it works for this demo. Users who clone the template would have to configure the CSP based on other integrations which they opted in for. Yes, it's a little more hassle this way but I think it's better than whitelisting every analytics or comment provider.

> I preference would be to only add `giscus.app` to the CSP so that it works for this demo. Users who clone the template would have to configure the CSP based on other integrations which they opted in for. Yes, it's a little more hassle this way but I think it's better than whitelisting every analytics or comment provider.

I agree with you, although this will affect the user experience, because we are hard to dynamically generate CSP, so we agree to only add `giscus.app`

timlrx:
Yes, this was also discussed in the [issue](https://github.com/timlrx/tailwind-nextjs-starter-blog/issues/304) by Alexander. There's a bit of trade-off between having a better default security practice vs ease of setting up the template.

timlrx (approved):

PR #311, opened 2021-12-26
Update README.md

lanbau:


timlrx:
Thanks - there's actually another easier way to embed a youtube video. From the youtube video, click on share, followed by embed. This would give you a iframe code which you can paste directly to your post. It will look something like this:

```
<iframe width="560" height="315" src="https://www.youtube.com/embed/-WyPBKKAL1Q?start=979" title="YouTube video player" frameBorder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowFullScreen></iframe>
```

You might need to use camel case for the `frameBorder` and `allowFullScreen` props, but otherwise that's all that is required.

timlrx:
Adding an external library to mount an iframe feels like an overkill for me. There's also mdx-embed which supports a large variety of social components.

lanbau:
thanks @timlrx that's a great point. Here's an update https://irvin.dev/blog/nextjs-tailwind-personal-blog. :smile:

PR #314, opened 2021-12-27
make project images optional with height adjustment

lanbau:


timlrx:
I pushed a modified version of your code in #315. Did not want to merge in a new package.lock and coded the conditional statement a little more cleanly. Thanks!

PR #322, opened 2022-01-04
fix: should use visibility to control display state of scroll top button

llovvoll:
The current version use "opacity" to control the display state of the button, but when it is hidden, the mouse can still click on the button in the lower right corner


timlrx:
Thanks, I think we can use `display` rather than `visibility` and use the tailwind classes rather than style?

llovvoll:
> Thanks, I think we can use `display` rather than `visibility` and use the tailwind classes rather than style?

ok~

PR #326, opened 2022-01-05
added my blog in the list :))

volcareso:
my blog is now in the list 
I am using this template to make it
Thanks a lot for this template !!!


volcareso:
Made improvements in there
- shortened the URL (omitted /blog/**/) the url is now like [url]/[post]

timlrx:
Thanks, for the submission. Would prefer to retain the current path structure.

PR #332, opened 2022-01-10
Add PWA support

karuppusamy-d:
## Changes:
Added Progressive Web App support using `next-pwa` and updated favicons and `site.webmanifest` to support PWA.

## Install Popup:

image

 
## PWA shortcuts:

image


next.config.js
```diff
@@ -52,49 +53,57 @@ const securityHeaders = [
...
+        })
+      }
+
+      return config
+    },
+
+    pwa: {
+      dest: 'public',
+      disable: process.env.NODE_ENV !== 'production',
+      publicExcludes: ['!sitemap.json', '!feed.json'],
```
AlexanderZeilmann, line 106: Shouldn't this be `publicExcludes: ['!sitemap.xml', '!feed.xml'],`?

karuppusamy-d: It excludes sitemap.xml and feed.xml from downloading automatically

Without `publicExcludes: ['!sitemap.json', '!feed.json']` workbox will download sitemap.xml and feed.xml automatically

image

AlexanderZeilmann: Yes, but why is the extension `.json` when you want to exclude `.xml`?

timlrx:
May I know why there's a need to replace the existing favicons?

I am also not sure about the benefits of making it PWA by default. Maybe we can keep it as a recipe for the next version of the template or temporarily put it in its own branch for reference. Happy to hear more opinions about this as well since I am not too familiar with PWAs.

next.config.js
karuppusamy-d: I fixed it now 👍🏻

public/static/favicons/site.webmanifest
```diff
@@ -1,14 +1,72 @@
...
+      "short_name": "Home",
+      "url": "/",
+      "description": "Home page",
+      "icons": [
+        {
+          "src": "/static/favicons/android-chrome-192x192.png",
+          "sizes": "192x192"
+        }
+      ]
+    },
```
AlexanderZeilmann, line 34: I would leave the `Open Home` shortcut away, because the link to the home page is automatically added and Chrome on mobile can only display 3 shortcuts.

karuppusamy-d:
> May I know why there's a need to replace the existing favicons?


As per [Google Developers](https://web.dev/installable-manifest/#recommendations)

PWA requires a 192x192px and a 512x512px icons.

For iOS icons should be: 180x180 pixels or 192x192 pixels.

AlexanderZeilmann:
I think the idea PWA is generally interesting (offline support, ...), but I am not sure whether this should be the default for blogs.

A concern I have is that browsers like Chrome on Mobile automatically display a notification that the website can be installed. This might be annoying to users. [There is a way to prevent this](https://stackoverflow.com/a/57557278), maybe it should be included here.

public/static/favicons/site.webmanifest
karuppusamy-d: Thanks for your suggestion, I will change that on the next commit.

rsipakov:
It's interesting, but if the project is for a single purpose as a blog, then PWA makes no sense. If the blog will be combined with other options of the application (for example SaaS project with blog), I think it be very useful and easy. I would agree with @timlrx to have a separate branch for the template, like for example with i18n. In this case, maybe in the future will add other futures related to the App like next-auth and etc. Also, as I see in the example [Xin chào](https://github.com/rsipakov/adebayosegun.com) lists, there is one option with the function PWA.

llovvoll:
Interesting PR, but I don't think the project need PWA, maybe try this in extra branch

thinkjrs:
@rsipakov It is easy enough for me to add PWA support separately to our blogs and I'm definitely in the use-case scenario you've outlined with multiple SaaS products and blogs. 

But rather than package this into the template, **I think the template is better without default PWA support**. 

One heuristic justification: using the default template media (favicons, etc...) for multiple blog sites is strange for users of an "app". 

Imagine that I like to read both timlrx's blog and my blog. So I install both of them: and now I have two of the exact same icons on my phone! As a developer, I would have already known not to do this and update the favicon and other media as required. And now the "usefulness" of PWA support out of the box is somewhat diminished. Furthermore, developers would need to update PWA defaults (which could be in or taken from the config for this blog) but still, it seems like too much for the default template.

And +1 on the extra branch. That said, unless there's just one "kitchen sink" branch this could add maintenance burdens on core devs.  

I do think we could include documentation instructions on quickly adding PWA support and would be happy to take on that PR over the coming weeks.

rsipakov:
> @rsipakov It is easy enough for me to add PWA support separately to our blogs and I'm definitely in the use-case scenario you've outlined with multiple SaaS products and blogs.
> 
> But rather than package this into the template, **I think the template is better without default PWA support**, too.
> 
> One heuristic justification: using the default template media (favicons, etc...) for multiple blog sites is strange for users of an "app".
> 
> Imagine that I like to read both timlrx's blog and my blog. So I install both of them: and now I have two of the exact same icons on my phone! As a developer, I would have already known not to do this and update the favicon and other media as required. And now the "usefulness" of PWA support out of the box is somewhat diminished. Furthermore, developers would need to update PWA defaults (which could be in or taken from the config for this blog) but still, it seems like too much for the default template.
> 
> And +1 on the extra branch. That said, unless there's just one "kitchen sink" branch this could add maintenance burdens on core devs.
> 
> I do think we could include documentation instructions on quickly adding PWA support and would be happy to take on that PR over the coming weeks.

@thinkjrs Thank you for your detailed comment. But I meant the same thing, I just put it in shorthand. You may have mistakenly tagged me and your comment was addressed to another participant )

timlrx:
Thanks all for the comments and @karuppusamy-d for the PR. Will close this branch for now but am happy to add your blog as a PWA example. 

After understanding the issue a little better, I don't think there's a need to include docs about PWA as well. Users who are interested in such a feature should be able to add the `next-pwa` plugin and make the necessary modifications themeselves.

PR #333, opened 2022-01-11
Add rehype-preset-minify

AlexanderZeilmann:
This allows the minification of blog posts.

### DIFF
--- a/lib/mdx.js
+++ b/lib/mdx.js
@@ -18,6 +18,7 @@ import rehypeAutolinkHeadings from 'rehype-autolink-headings'
 import rehypeKatex from 'rehype-katex'
 import rehypeCitation from 'rehype-citation'
 import rehypePrismPlus from 'rehype-prism-plus'
+import rehypePresetMinify from 'rehype-preset-minify'
 
 const root = process.cwd()
 
@@ -83,6 +84,7 @@ export async function getFileBySlug(type, slug) {
           { bibliography: frontmatter?.bibliography, path: path.join(root, 'data') },
         ],
         [rehypePrismPlus, { ignoreMissing: true }],
+        rehypePresetMinify,
       ]
       return options
     },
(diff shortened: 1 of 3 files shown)


timlrx:
There's no need for the plugin as both esbuild which is used by mdx-bundler and webpack / swc from nextjs will take care of it. It's intended for use in standalone applications.

AlexanderZeilmann:
Adding the plugin does have an effect (-> right-click -> view source):
* [source of a blog post **without** the plugin](https://tailwind-nextjs-starter-blog.vercel.app/blog/the-time-machine)
* [source of a blog post **with** the plugin](https://tailwind-nextjs-starter-blog-beta.vercel.app/blog/the-time-machine)

esbuild / webpack / swc do minify the page but not the blog post.
It would probably make more sense to configure these build tools such that the blog post (and e.g. the `json` starting in line 2) is also minified. However, I haven't looked into the build setup too closely to fix this.

timlrx:
Thanks, good to know. Can we try turning on the minify option in `esbuild`? Mdx-bundler has an example over [here](https://github.com/kentcdodds/mdx-bundler#esbuildoptions). Would prefer this approach over a rehype plugin if this works.

AlexanderZeilmann:
I'm not seeing any minification when using `options.minify = true` in the `esbuildOptions`.

timlrx:
Took a deeper look into this issue. Mdx-bundler sets [minify to true by default](https://github.com/kentcdodds/mdx-bundler/blob/main/src/index.js#L202). Changing it to false does show a difference. 

The output from xdm is probably treated by esbuild as a string literal and it does not minify it. Hence the plugin is required, if we truly want to minimize the size of HTML over the wire. It seems to be relatively minor about 6% smaller and am not too sure how much longer it takes to process but happy to merge in the change. Thanks!

timlrx (approved):

PR #335, opened 2022-01-14
Remove script format from "or" text by removing indent

localsite:
Also changed text to: "or with TypeScript" rather then "TS"


timlrx (approved):
Thanks!

PR #338, opened 2022-01-16
Add  *.googletagmanager.com to script-src

wozniaklukasz:
After adding `googleAnalyticsId` analytics did not work because the google script was blocked.


timlrx:
I decided not to whitelist all the additional plugins because it depends on what services you opt-in for. Maybe I should add it in the readme...

PR #343, opened 2022-01-18
fix: incorrect url generated for authors pages in sitemap

burhanuday:
previously the authors urls in the sitemap would be generated as
```
sitename.comdata/authors/default
```
this would make Google's SEO dashboard thrown an error since it would consider the url outside of the current submitted domain

this change fixes the issue in the same manner as how blog urls are for sitemaps. have confirmed and tested the change

### DIFF
--- a/scripts/generate-sitemap.js
+++ b/scripts/generate-sitemap.js
@@ -22,6 +22,7 @@ const siteMetadata = require('../data/siteMetadata')
                 const path = page
                   .replace('pages/', '/')
                   .replace('data/blog', '/blog')
+                  .replace('data/authors', '/authors')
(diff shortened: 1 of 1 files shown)


burhanuday:
@timlrx please review

timlrx:
That's because there are [no author pages supported by the template](https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/master/scripts/generate-sitemap.js#L9-L14). If a user chooses to add a multi-author page then you will need to whilelist that path route and add the line in your PR. 

There are other possibilities including displaying all the authors in a single page, which the current strategy works. In general, I do not assume more than is required so the current configuration works fine.

PR #346, opened 2022-01-19
refactor: missing a value for `letterSpacing`

rsipakov:
If used like this definition `letterSpacing: theme('letterSpacing.tight')`, we need to have its value or remove it from the code.

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -6,6 +6,9 @@ module.exports = {
   darkMode: 'class',
   theme: {
     extend: {
+      letterSpacing: {
+				tight: "-.05em",
+			},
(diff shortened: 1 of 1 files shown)


timlrx:
It's a default, see [tailwind docs](https://tailwindcss.com/docs/letter-spacing#customizing-your-theme).

rsipakov:
According to this docs, the default value `tightest` not `tight`

> It's a default, see [tailwind docs](https://tailwindcss.com/docs/letter-spacing#customizing-your-theme).

Right. But for some reason, it doesn't always work (on my side) for MDX without specifying it in tailwind.config.js

PR #354, opened 2022-01-24
Create `data/blog` directory if missing

edwincoronado:
If for any reason the `data/blog` directory is not present, the script fails. This change creates the `data/blog` directory if it doesn't yet exist to prevent error.


timlrx (approved):
LGTM, thanks!

PR #359, opened 2022-01-26
fix: insert code title relative to parent

timlrx:
Fix https://github.com/timlrx/rehype-prism-plus/issues/29

Code title should be inserted relative to the parent, not the root component as it may be a child itself e.g.

````
<div style={{height: '300px', overflow: 'scroll'}}>
```json:fileName.json
A very long code snippet
```
</div>
````


timlrx:
@AlexanderZeilmann this should fix the issue.

AlexanderZeilmann:
I can confirm this fixes it.

Issue #366, opened 2022-01-31
Draft articles are included in the sitemap


iamcam:
**Describe the bug**
When the sitemap is generated, drafts are included (they shouldn't be).

**To Reproduce**
1. Create a draft page/entry
2. generate the site
3. inspect the sitemap
4. observe the generated sitemap includes the draft pages

**Expected behavior**
Drafts should be excluded

**System Info (if dev / build issue):**
- next 12.0.7
- yarn 1.22.4
- node 16.3.0
-
**Browser Info (if display / formatting issue):**
N/A

PR #367, opened 2022-01-31
Exclude drafts from the sitemap, added typescript file to glob

iamcam:
This PR...

1. Excludes drafts from the sitemap
2. Includes typescript pages

thus, resolving bug https://github.com/timlrx/tailwind-nextjs-starter-blog/issues/366

### DIFF
--- a/scripts/generate-sitemap.js
+++ b/scripts/generate-sitemap.js
@@ -1,5 +1,6 @@
 const fs = require('fs')
 const globby = require('globby')
+const matter = require('gray-matter')
 const prettier = require('prettier')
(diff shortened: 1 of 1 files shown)


timlrx:
Thanks, I am refactoring pieces over to `contentlayer` to avoid consistently parsing the markdown file, but this does the trick for now.

iamcam:
Lee Robinson uses `content layer`, and an early revision of my new site was based off his work back in December. Maybe I had done something wrong, but I found `contentlayer` a bit clumsy, though the premise is good.

PR #384, opened 2022-02-11
pin minor typescript version

erikdstock:
Forking this repo and using a different package manager which does not respect the existing package-lock file will cause ts 4.5 to be installed, which [hangs during build](https://github.com/microsoft/TypeScript/issues/46900).


timlrx:
Thanks, this is an irritating bug. Let's pin it to the minor version until it is fixed.

PR #393, opened 2022-02-15
added keyboard buttons feature to show sequence of buttons

404answernotfound:
Hello, I am using this nextjs starter for my blog and I added the possibility to use graphic buttons sequences on posts depending on a frontmatter attribute. Would you consider accepting my pull request for this small feature?

### DIFF
--- a/components/SEO.js
+++ b/components/SEO.js
@@ -72,6 +72,7 @@ export const TagSEO = ({ title, description }) => {
 }
 
 export const BlogSEO = ({
+  keyboard,
   authorDetails,
   title,
   summary,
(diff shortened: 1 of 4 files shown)


timlrx:
The code looks a little strange to me as you are highjacking the SEO component to render the link tag. In any case, I think it would be best if you implement it on your blog and add the example to the demo blog list since it is a niche feature (in my opinion). Other users that might want to emulate it could check out your blog.

404answernotfound:
> The code looks a little strange to me as you are highjacking the SEO component to render the link tag. In any case, I think it would be best if you implement it on your blog and add the example to the demo blog list since it is a niche feature (in my opinion). Other users that might want to emulate it could check out your blog.

That's also a nice idea, thanks! Also, keep up the great work!

PR #427, opened 2022-03-27
Render floating comment button only if a provider is specified

cdransf:
This pull request updates the `ScrollTopAndComment.js` component to only render the comment button in cases where a comment provider is specified at `siteMetadata.comment.provider`.

Granted, the theme default is to specify a provider, but, in cases where that provider is cleared and set to an empty string, the comment button still renders but directs users to a commenting implementation that isn't rendered.


timlrx:
Makes sense to me, thanks!

PR #450, opened 2022-05-02
set viewport to fix text scaling issue in nextjs and tailwind

christiaanNagel:
> Hmm could not find what's wrong with this. It works when I change the font size for my desktop but not for my android chrome browser. If anyone knows what is the issue, would appreciate a PR, thanks!

Solution : Add the head component and set the default viewport

```javascript
import Head from "next/head";
<Head>
  <meta name="viewport" content="width=device-width, initial-scale=1.0"></meta>
</Head>
```

[link to forum discussion](https://www.reddit.com/r/tailwindcss/comments/kj146o/i_am_struggling_with_tailwinds_responsive_design/)

### DIFF
--- a/pages/index.js
+++ b/pages/index.js
@@ -4,6 +4,7 @@ import Tag from '@/components/Tag'
 import siteMetadata from '@/data/siteMetadata'
 import { getAllFilesFrontMatter } from '@/lib/mdx'
 import formatDate from '@/lib/utils/formatDate'
+import Head from "next/head"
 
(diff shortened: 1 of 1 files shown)


timlrx:
Prettier is throwing an error. Do you mind running it through the linter before submitting a PR? Thanks.

Also, the [default viewport has already been included in _app.js](https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/master/pages/_app.js#L21-L23) and would be injected in all pages. I don't see why adding it to only index.js would fix the issue.

PR #456, opened 2022-05-11
Add rehashly.com to the examples list

hgad:
Add rehashly.com to the list of example websites using this template.


timlrx:
This seems like linking to a commercial website for SEO points...

PR #465, opened 2022-05-26
Fix: Passing Sorted Blogs to ListLayout On [page].tsx

reubence:
Hey Tim, I've encountered a small bug which was passing the un-sorted blogs to the ListLayout on the `./blog/[page].tsx` page. I've added the `sortedBlogPost()` to `getStaticProps()` so that the sorted blog posts are passed to the ListLayout Component. 

It's my first time contributing, do let me know if this requires any further clarification. Thanks for all your hard work you've put in to bring us this template!


timlrx:
Good find, thanks for the PR!

PR #470, opened 2022-05-28
fixed glitch in mobile navbar

thekamalkashyap:
when we scroll the  page and try to open sidebar, their is some gap between the navbar and top of the sidebar which can spoil user experience.
Screenshot_2022-05-28-21-43-40-53_40deb401b9ffe8e1df2f1cc5ba480b12
i converted sidebar to fullscreen and added close sidebar button inside itself.


timlrx:
Sure, we can go with this instead, thanks!

PR #478, opened 2022-06-10
frontmatter images can now be external

mlevit:
BlogSEO images were hardcoded to be locally sourced only. This change performs a simple check (i.e., `http` present in URI).


timlrx:
Thanks for the fix!

Issue #474, opened 2022-06-07
`theme-color` should use `prefers-color-scheme` instead of fixed color


galligan:
**Is your feature request related to a problem? Please describe.**
Currently the `<meta` `theme-color` uses `content="#000000"` which obviously means that the blog will always have a black background surrounding a notch on phones that have them.

**Describe the solution you'd like**
Instead of using a fixed color for all states, the `theme-color` meta should reference the preferred color scheme…such as:

``` html
<meta name="theme-color" media="(prefers-color-scheme: light)" content="#fff" />
<meta name="theme-color" media="(prefers-color-scheme: dark)" content="#000" />
```

**Additional context**
Something nice would be to make it possible to use the same background color as referenced in `body`. Though that may need to be abstracted into a different component to make that happen.

PR #482, opened 2022-06-15
feat: use prefers-color-scheme for meta theme-color

timlrx:
Fic #474


galligan:
Very nice @timlrx. Confirmed working on my iPhone!

PR #493, opened 2022-06-27
add PostHog analytics - The product analytics suite you can host yourself

anuuyu:
PostHog is the open source platform for your product data, complete with event pipelines, a data warehouse, and an API. 
PostHog analytic include free tier up to 1 million events or use self-hosting.  learn more https://posthog.com


timlrx:
Thanks, I will probably refactor it to make `api_host` an argument, but otherwise looks good to me.

PR #499, opened 2022-07-05
Fix URL to Timothy Lin's website in README file

KuraFire:


timlrx:
Good catch, thanks :)

PR #506, opened 2022-07-18
Upgrade glob-parent to fix security issue

josemiguel-alvarez:
### Problem
glob-parent before 6.0.1 is vulnerable to Regular Expression Denial of Service (ReDoS).

The security issue was fixed https://github.com/gulpjs/glob-parent/pull/49.

### Suggested solution
Require version ^6.0.1 of glob-parent using NPM overrides in package.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14800 @@
 {
   "name": "tailwind-nextjs-starter-blog",
   "version": "1.5.6",
-  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


timlrx:
Hi, thanks for raising this but I think there's no need for the patch. It's used internally only by tailwind, contentlayer, eslint and globby. No user is ever going to submit an input for parsing.

PR #512, opened 2022-07-23
Remove sharp dependency

josemiguel-alvarez:
sharp is not being used and it is related to 2 vulnerability issues. To keep code clean and solve this security issues it's better to just remove it from the dependencies list.

- https://github.com/josemiguel-alvarez/tailwind-nextjs-starter-blog/security/dependabot/6
- https://github.com/josemiguel-alvarez/tailwind-nextjs-starter-blog/security/dependabot/2

This PR also updates the lockFileVersion to v2. v2 is the currently supported version in the current Node LTS version. v2 is backwards compatible with v1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,14399 @@
 {
   "name": "tailwind-nextjs-starter-blog",
   "version": "1.5.6",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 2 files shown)


timlrx:
sharp is recommended if you are hosting your own nextjs application. It's what vercel uses in production build as well. I might remove it for other reason - just to keep the download size smaller and those who want to self-host can install it themselves, but that will probably be for the next version release under [pliny](https://github.com/timlrx/pliny). I will probably not touch most of the dependencies in the current version.

PR #516, opened 2022-08-02
Fix wrong option name passed to bundleMDX

sabigara:
I believe this is a typo

https://github.com/kentcdodds/mdx-bundler#mdxoptions


PsyGik:
Not a typo. See https://github.com/kentcdodds/mdx-bundler/releases/tag/v9.0.0

> We now use @mdx-js/esbuild instead of xdm and xdmOptions has been renamed to mdxOptions

PR #519, opened 2022-08-06
Upgrade several node dependencies

bh:
- The `output:standalone` [feature of NextJS](https://nextjs.org/docs/advanced-features/output-file-tracing) works fine; my website has a size of 13 MB including all the node modules.
I built the standalone version with this Circle CI snippet here:
```bash
      - run: 
          name: "Build NextJS"
          command: |
            yarn build
            mkdir dist
            mkdir -p app/standalone/public
            cp ./package.json app
            cp -r .next/standalone app
            cp .next/static app/standalone/.next/ -r
            cp -r public/* app/standalone/public/
            cp ecosystem.config.js app
            tar -cJf dist/standalone-dist.tar.xz app   
      - persist_to_workspace:
          root: "."
          paths:
            - ./dist/standalone-dist.tar.xz
```

- Upgrade of globby needs to convert the generate-sitemap.**js** script to generate-sitemap.**mjs**
- I had no success without the preact compat plugin I have introduced; without it, the Next JS build has a load of 100% CPU, I had no success to debug this.


timlrx:
Would prefer not to update the dependencies as I am working on another major version update to this. I guess the additional build commands might be because you are hosting locally. Thanks, for the PR but I will close this for now.

PR #528, opened 2022-08-19
Create enclosing folder in the title in compose script.

moojing:
# Summary
add some logic to let user create subfolder (or sub route) when running the compose script.

# Screenshot 
Aug-19-2022 22-13-01

### DIFF
--- a/scripts/compose.js
+++ b/scripts/compose.js
@@ -100,15 +100,21 @@ inquirer
     },
   ])
   .then((answers) => {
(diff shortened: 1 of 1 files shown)


timlrx:
It's a nice feature improvement but the next version that I am rolling out is not going to support the compose script. Thanks for the PR but I don't think I will merging it in.

moojing:
It's okay, it's just an idea that suddenly hit me.
BTW this is really a good tool for making a blog site, thank you 😌.

PR #547, opened 2022-09-21
fix mdx.js configuration typo

benschac:
per the documentation: https://github.com/kentcdodds/mdx-bundler#mdxoptions

This configuration has a typo in it.

### DIFF
--- a/lib/mdx.js
+++ b/lib/mdx.js
@@ -60,7 +60,7 @@ export async function getFileBySlug(type, slug) {
     source,
     // mdx imports can be automatically source from the components directory
     cwd: path.join(root, 'components'),
-    xdmOptions(options, frontmatter) {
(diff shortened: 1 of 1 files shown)


lib/mdx.js
```diff
@@ -60,7 +60,7 @@ export async function getFileBySlug(type, slug) {
     source,
     // mdx imports can be automatically source from the components directory
     cwd: path.join(root, 'components'),
-    xdmOptions(options, frontmatter) {
+    mdxOptions(options, frontmatter) {
```
yannbf, line 63: Maybe it changed in newer versions, but the `xdmOptions` option is actually correct. 
image

timlrx:
Will be migrating to a new setup with contentlayer in v2

PR #550, opened 2022-09-29
Updated html language to sitemetadata and added new example website

qmau94:
Hey @timlrx, thanks for the awesome template ❤️

As the HTML lang is 'en' by default so I made a change to ensure it respects the site metadata config.

By the way, I also added [our site](https://dev.vntechies.com/) using your template. Big thanks!


timlrx:
Thanks, I will make the changes on the Pliny repo. Holding off from listing other websites for now until I have launched v2.

PR #553, opened 2022-10-05
Modify configuration for GCP deployment

shottah:
## Description

The deployment to GCP uses the `gcloud app deploy` script which compiles the `next.config.js` at runtime, however the  `@next/bundle-analyzer` is a `devDependency` so the deployment stage will fail.

## Drive-by changes

Added instructions for deploying to GCP (Google App Engine) to `README.md`


timlrx:
Thanks for pointing that out, will also copy the changes to the pliny repo for v2

PR #565, opened 2022-11-04
docs: update Lighthouse report's link

lauthieb:
Hello,
Thanks for this template!
I propose to update the Lighthouse report's test result.

## Before

CleanShot 2022-11-04 at 08 47 22

## After

CleanShot 2022-11-04 at 08 47 29

Thanks :)


timlrx:
Thanks for the update, I probably want to redo the benchmark for v2 of the template with possibly nextjs 13, but this will work for now.

Issue #595, opened 2023-01-13
author layout problem


MohammadTaseenKhan:
when there is more than two authors the layout becomes unresponsive to the width

screenshot:

![image](https://user-images.githubusercontent.com/85288261/212346512-a787dadf-5667-4ae3-b212-3c54146f7ce5.png)
![image](https://user-images.githubusercontent.com/85288261/212346561-c6131b0d-11d1-487f-a424-d1ab7bb6bef7.png)

@timlrx please help

timlrx:
Thanks, I have fixed it. If you want to update the style you can just update to this line: https://github.com/timlrx/pliny/blob/main/starter-blog/layouts/PostLayout.tsx#L65

PR #596, opened 2023-01-15
updated the author section overflow

MohammadTaseenKhan:


MohammadTaseenKhan:
@timlrx i have updated the author section by adding overflow

image
image

the cause is already available in the issue section
https://github.com/timlrx/tailwind-nextjs-starter-blog/issues/595

MohammadTaseenKhan (commented):
update author section

PR #598, opened 2023-01-23
fix: handle posts with missing tags

jurnalanas:
This PR addresses an issue where the /tags page would return an error if a markdown post file was missing the `tags` field in the frontmatter. 

Screenshot 2023-01-24 at 06 25 10


timlrx:
Will be cleaning things up and migrating to a v2 version with react server components based on the app dir structure.

PR #619, opened 2023-03-11
Fix giscus width issue.

sjwaight:
When comments load they only take 25% of screen. This CSS changes fixes the issue.


adrianfuro:
I can confirm that is working !

jonade:
Related to #624 . The CSP change fixes this without needing to make CSS changes

sjwaight:
CSP change is the correct fix for this, so I'm closing this PR.

PR #621, opened 2023-03-13
Add quotes to title and summary

silent1mezzo:
Posts and summaries can often have punctuation that breaks markdown parsing without wrapping them in quotes


silent1mezzo:
@timlrx anything needed to get this merged?

timlrx:
@silent1mezzo sorry for the late reply. The fix looks good to me. I will merge it in and this will probably be the last patch to v1! I do not plan on adding the compose script in v2 as I find it easier to just copy an existing markdown file to get started and the new contentlayer schema and typings make the frontmatter fields quite obvious.

silent1mezzo:
Ah interesting, we use it exclusively since it's an easy process to follow. Are there any docs on migrating to v2? I upgraded to nextjs13 but hadn't pushed upstream yet.

timlrx:
If you did not make many changes to the previous template, it might be easier to just clone the new version and copy your content over there. 

There's quite a bit of changes going from the page directory to app directory setup, from changes in the behaviour of core components like `next/link` and using `next/navigation` instead of `next/router` to using the recommended best practices like the new metadata export instead of rolling our own SEO header.

I have also abstracted a lot of component logic to a new library - [pliny](https://github.com/timlrx/pliny), so they can be more easily reused across different sites and templates. Again this primary supports the new next 13 with app directory setup. Unfortunately, it's too much work to support both app and page directory setup...

More detailed changes and updates can be found in the v2 branch and [pull request](https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/658). I plan on writing a more detailed post outlining how to migrate content while maintaining the look and feel of v1 tomorrow, before finally merging it. If you have made significant changes, it might just be easier to continue with the pages approach and cherry pick changes that you like from the new version.

PR #656, opened 2023-07-15
update sharp version

itoffshore:
* fixes moderate vulnerability from:

  - https://github.com/advisories/GHSA-gp95-ppv5-3jc5


timlrx:
The new version no longer installs sharp by default which follows nextjs recommendations and users can do so if they wish to self host.

PR #657, opened 2023-07-16
fix: allBlogs filter

yunsii:


timlrx:
No longer using a custom function to generate sitemap since we can do it with the in build next 13 sitemap support.

PR #674, opened 2023-08-09
Add missing directory to lint command

petrsiegl:
Add new `app` directory of next.js 13 for lint. Keeping `pages` for now as they can coexist as people upgrade


timlrx:
Thanks, good catch.

PR #681, opened 2023-08-15
Integrate Preact Library

tuanductran:
In this pull request, I have integrated the Preact library into your project to enhance performance and optimize the source code size of the application. Preact is a lightweight JavaScript library with a syntax similar to React, but it has a smaller footprint and higher performance optimization.

Before utilizing the Preact library.
Screen Shot 2023-08-15 at 16 34 12

After incorporating the Preact library.
Screen Shot 2023-08-15 at 16 37 49

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -70,6 +70,7 @@ module.exports = () => {
     },
     experimental: {
       appDir: true,
+      esmExternals: false,
     },
     async headers() {
       return [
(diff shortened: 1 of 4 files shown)


timlrx (dismissed):

timlrx:
This is error prone since the preact has not support react server components and there might be a mismatch between the server and client generated code.

There were previous issues with package compatibility as well e.g.The search functionality does not seem to be working on the preview. I think it's better to use react as the default and let users customize it to use preact if they understand the issues around it.

tuanductran:
Okey

PR #692, opened 2023-08-23
fix non-english tags filter failed, display error

chasingboy95:
when tag is not English, the Tag&Blog tab can not match the right blogs, and the title of web site display not correct
image


timlrx (approved):
Thanks for the fix

PR #700, opened 2023-08-27
fix: hardcoded check for draft in development

rareyesdev:
`allCoreContent` checks for the draft state based on NODE_ENV which is the correct behavior. This check is redundant and is causing the tags page to skip drafts in development.


timlrx (approved):
LGTM, thanks!

PR #702, opened 2023-08-27
fix: pass components in about page

rareyesdev:
The MDXLayourRenderer needs to receive the `components` prop in order to render some elements correctly.

For example, external links will not be rendered with the correct anchor attributes.

## Before
image

## After
image


timlrx:
This was actually intentionally as I wanted to show that different pages could have different components.

PR #708, opened 2023-09-03
fix: sort posts in tags page

L4ys:
In the `/tags/<tag>` page, the posts were not sorted, so I added a `sortPosts` function call


timlrx:
There's some prettier formatting error. Could you run it through prettier (which should be triggered automatically as a pre-commit hook) before re-commiting?

timlrx:
Thanks for the fix!

iliamrv:
> In the `/tags/<tag>` page, the posts were not sorted, so I added a `sortPosts` function call

L4ys, do you have idea how to integrate metatags to the Tags page? I am looking for such functionality.

PR #715, opened 2023-09-22
bugfix: Fix for missing image in ld+json blob

dishs:
Fix for the missing image in the JSON/LD structured data.

After fix:
image


timlrx (approved):
Thanks, good catch. It appears that contentlayer is not able to parse list of strings properly in the computed fields and needs it to be json. The same problem applies to the `author` field as well - I will fix that in another PR.

PR #733, opened 2023-10-11
fix: use sort posts for rss, fix lastBuildDate

ttys3:
as generateRSS get last build date from first post, so we need to use sorted posts, otherwise the `lastBuildDate` value is wrong


ttys3:
@timlrx

timlrx (approved):
Thanks for the fix!

PR #747, opened 2023-10-28
fix: filter draft posts from home & blog pages

billyjacoby:
Before this change a post with a draft status of `true` was still showing on the home and blog pages.

### DIFF
--- a/app/blog/page.tsx
+++ b/app/blog/page.tsx
@@ -8,7 +8,7 @@ const POSTS_PER_PAGE = 5
 export const metadata = genPageMetadata({ title: 'Blog' })
 
 export default function BlogPage() {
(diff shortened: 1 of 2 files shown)


timlrx:
It is already filtered out in the `allCoreContent` function. It shows only in dev / preview but not in prod. You can verify that the deployed website does not have this post: https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/main/data/blog/my-fancy-title.mdx

billyjacoby:
Ahhhh got it, that makes sense! I didn't see that in the docs - i can add that if you'd like!

timlrx:
Feel free to create a new PR if you would like to add to the docs, thanks!

PR #758, opened 2023-11-06
Closes #757.  Added return notFound() to app/blog/[...slug]/page.tsx …

### LINKED ISSUE #757: Dynamic blog post missing doesn't return 404 but returns under construction
**Describe the bug**
Whenever you navigate to a route that a blog post doesn't exist, i.e. http://localhost:3000/blog/does-not-exist then you'll receive a 200 with an under construction page.  It should return a 404 since it doesn't exist otherwise Google will continue to index the page.

Solution:
https://nextjs.org/docs/app/api-reference/functions/not-found

Simply import notFound from next/navigation and return that instead of the construction react component.

```
import { notFound } from 'next/navigation'
if (postIndex === -1) {
    return notFound()
  }
```

Romstar:
…instead of under construction so that missing pages return 404s and default error page


timlrx (approved):
Thanks for the fix!

PR #771, opened 2023-11-12
feat: support chinese tag name

z2015:
For chinese tag name in devlopment mode will get error below:

_Error: Page "/tags/[tag]/page" is missing param "/tags/%E6%97%A5%E8%AE%B0" in "generateStaticParams()", which is required with "output: export" config._

Should be fixed with encodeURI method.

### DIFF
--- a/app/tags/[tag]/page.tsx
+++ b/app/tags/[tag]/page.tsx
@@ -25,7 +25,7 @@ export const generateStaticParams = async () => {
   const tagCounts = tagData as Record<string, number>
   const tagKeys = Object.keys(tagCounts)
   const paths = tagKeys.map((tag) => ({
-    tag: tag,
+    tag: encodeURI(tag),
   }))
   return paths
 }


timlrx (approved):
Thanks for the fix!

klaaay:
I noticed that with this modification, the "tag" in ﻿params will be encoded twice in the production environment. Thus, it needs to be decoded twice to restore its normal Chinese display.

```diff
export default function TagPage({ params }: { params: { tag: string } }) {
+   const tag = decodeURI(decodeURI(params.tag))
  // Capitalize first letter and convert space to dash
  const title = tag[0].toUpperCase() + tag.split(' ').join('-').slice(1)
  const filteredPosts = allCoreContent(
    sortPosts(allBlogs.filter((post) => post.tags && post.tags.map((t) => slug(t)).includes(tag)))
  )
  return <ListLayout posts={filteredPosts} title={title} />
}
```

PR #772, opened 2023-11-12
Replaced blue tailwind class with primary theme class

chrisvasey:
I noticed the 404 page still uses the Tailwind CSS blue colour for its button.
Replaced with primary class.


timlrx:
I just like using blue for this, but you could replace it with primary for your blog

PR #778, opened 2023-11-17
draft posts are not included in the sitemap

zziri:
I am creating this PR because I think posts that are in draft status should not be added to the sitemap.
I would appreciate your feedback. 🙏


timlrx:
Good catch, yes you are right that the draft post currently shows up in sitemap.

PR #798, opened 2023-12-09
docs: remove Terminals.run demo project

patrickvaler:
[Terminals.run](https://terminals.run) is no longer available. Due to that it should be removed from the listed demo projects.


timlrx:
Seems like it is back up?

patrickvaler:
> Seems like it is back up?

Right, seems to be available again 👍

PR #801, opened 2023-12-11
Add type to `projectsData.ts`

iamkroot:
Without this, I get a build failure due to the TS checker.


timlrx:
The project has strict mode set to `false`. You might have enforced it. In any case, thanks for the enhancement!

PR #809, opened 2023-12-22
fix: run lint and add changelog file

johnnyvaz:
* I just ran npm run lint to avoid giving an initial error
* Add lib for build release
* Insert Version on footer
* Add changelog.md file


timlrx:
There's no need for an additonal change. I manage the change log and version through github.

PR #850, opened 2024-02-13
fix: layout shifting issue with navbar caused by scroll bar

iksflow:
There is an issue with layout shifting when clicking the navbar as shown in the image below. 
I resolved it by adding padding-left to the body tag.
Screen Recording 2024-02-13 at 3 23 53 PM


timlrx:
Thanks, this is a very clever hack!

PR #858, opened 2024-02-19
feat: Move conditional rendering inside svg to prevent layout shift

Spiteless:
While the `ThemeSwitch` mounts there's a potential for the buttons to be shifted to the left if it loads in slowly.

Moving the conditional rendering inside the `<svg/>` keeps the spacing open for the button to load in properly.


timlrx:
Thanks for the fix!

PR #866, opened 2024-03-04
Fix: Don't need to encode tags when production

leonliu06:
The tags with Chinese character would be double encoded when deploy in the way of static export. Then the server wold not  find the correct route.


timlrx:
Hmm I just noticed that there were some people who posted about this issue in https://github.com/timlrx/tailwind-nextjs-starter-blog/pull/771 as well. What's the reason that `encodeURI` is needed in dev but not prod? And could you just if it works on a normal `yarn build` and `yarn serve` as well?

Would like to fix this issue properly.

1226085293:
I also encountered this problem. I hope to merge as soon as possible

PR #886, opened 2024-04-07
Fixed Github Pages Typo

yyexela:
Fixed a typo I found when deploying the page to GH Pages.


timlrx:
Not a bug, I specify the absolute src path e.g. https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/test-gh-pages-static/data/blog/pictures-of-canada.mdx?plain=1#L23 so the concatenation is correct as seen in the deployed example. But, if you are inconsistent with the `/` then yes, you might need to modify it accordingly.

PR #888, opened 2024-04-08
chore: fix typo (wrong path)

umihico:
`data/img/ocean.jpg` appears to be unused throughout the project. All references are currently  `/static/images/ocean.jpg`.


timlrx:
Thanks for the fix

PR #890, opened 2024-04-15
Make navbar fixed

dipshikha-adhikari:
We had to scroll up to click links when reached to the bottom. Now the header is fixed and the problem is solved.


timlrx:
Not a bug, just a preference. There's even a button to go straight to the top for navigation purposes.

PR #906, opened 2024-04-30
Fix z-index in ThemeSwitch.tsx

PxlSyl:
Hi! Here's the PR for the smal bug with the post-banner layout!


timlrx:
Thanks for the fix!

PR #912, opened 2024-05-06
remove markpitblado.me from README v2

mark-pitblado:
Loved the theme but I have now switched to a different framework and thus no longer fair to claim space here!

Thanks for all your hard work!


timlrx:
Thanks for closing the loop - you're the first!

PR #940, opened 2024-06-16
feat: Add hidden to mobile nav

Spiteless:
## `Fix`

Added `hidden` class to `navShow` classes so that the MobileNav elements are removed from tab indexing order.

#### Expected behavior:
Pressing tab will navigate through visibile links, buttons

#### Current behavior before PR:
Pressing tab will switch from ThemeSwitch icon to MobileNav links, which are hidden on larger screens


timlrx:
Thanks, good catch. Run `prettier` and I will re-deploy the updated version.

Spiteless:
Looks like I just got `husky` and `prettier` working but they aren't applying any changes to this PR.

Spiteless:
This PR will be redundent if #943 is merged, closing

PR #943, opened 2024-06-19
Refactor MobileNav to use headlessui

Spiteless:
## `Enhancement`

Current mobile drawer doesn't take advantage of accessibility offered by `headlessui`

Used [`headlessui/Dialog`](https://headlessui.com/v1/react/dialog) as an example to refactor the `MobileNav` drawer

## Changes
 - Move close button to bottom of drawer HTML so it is not auto selected when entering `MobileNav`
 - Kept current `MobileNav` styling
 - Changed to `Dialog` as base component
   - Allows for tab press to stay within `Dialog`
   - Allows for esc press to close `Dialog`

## Visuals
<details>
<summary>Before Changes</summary>
image
image

</details>

<details>
<summary>After Changes</summary>

image
image

</details>

### DIFF
--- a/components/MobileNav.tsx
+++ b/components/MobileNav.tsx
@@ -37,7 +37,7 @@ const MobileNav = () => {
       </button>
       <div
         className={`fixed left-0 top-0 z-10 h-full w-full transform bg-white opacity-95 duration-300 ease-in-out dark:bg-gray-950 dark:opacity-[0.98] ${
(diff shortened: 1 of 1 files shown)


timlrx:
Resolved merge conflict by adding some of the changes from the previous PR. Otherwise the changes looks good to me. Now that we are using `headlessui` for the theme switch component, I think this change makes sense and the accessibility improvement is great.

PR #950, opened 2024-07-01
Fix button alignment in ThemeSwitch component

cxf213:
### Fix button alignment in ThemeSwitch component

This PR fixes the alignment issue of the button within the ThemeSwitch component. The button was not properly centered within its container, causing it to appear misaligned in the navigation bar. This fix ensures that the button is vertically centered.

#### Changes Made:
- Applied `flex` layout and alignment classes to the div containing `Menu.Button` in `ThemeSwitch.tsx`.

#### Before:
image

#### After:
image

### DIFF
--- a/components/ThemeSwitch.tsx
+++ b/components/ThemeSwitch.tsx
@@ -54,7 +54,7 @@ const ThemeSwitch = () => {
   useEffect(() => setMounted(true), [])
 
   return (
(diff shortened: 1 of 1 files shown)


timlrx:
Thanks for the fix! In the future, do run the code through prettier before committing.

PR #957, opened 2024-07-07
Allow line breaks for <a> tags to keep body width on mobile

umihico:
[My blog's article](https://umihi.co/blog/20221226-login-aws-console-by-cli-credentials) has URL references at the end, and it create extra right margin on entire mobile layout. I confirmed that adding `className="break-words"` on `<a>` tags fixed that.

| **Before** | **After** |
| ------------- | ------------- |
| スクリーンショット 2024-07-07 11 32 36 | スクリーンショット 2024-07-07 11 32 31 |

Thank you for sharing this blog template.


timlrx:
That makes sense, thanks for the fix! In the future, do run the code through prettier before committing.

PR #959, opened 2024-07-07
devcontainer

abernier:
will allow:
- open in GH Codespace
- open in VScode local devcontainer
- open in Stackblitz, CSB... every editor supporting devcontainers


timlrx:
Tested locally and works well for me.

PR #960, opened 2024-07-10
generator

abernier:
# 1. Generator

This PR adds a "[reusable](https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow) workflow" that any new {repo} can re-"use" as a generator(think Jekyll), eg:

```
on:
  push:
    branches: main

jobs:
  build:
    uses: timlrx/tailwind-nextjs-starter-blog/.github/workflows/action.yml@v2

...
```

The repo can override any file from `timlrx/tailwind-nextjs-starter-blog`, especially `data/blog/` folder.

-> A working example: [abernier/myblog](https://github.com/abernier/myblog)

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,60 @@
+name: Reusable tailwind-nextjs-starter-blog workflow
+
+on:
+  workflow_call:
+    inputs:
+      repo:
+        type: string
+        required: false
+        default: 'timlrx/tailwind-nextjs-starter-blog'
+      ref:
+        type: string
+        required: false
+        default: 'v2'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          repository: ${{ inputs.repo }}
+          ref: ${{ inputs.ref }}
+      - uses: actions/checkout@v4
+        with:
+          path: overrides
+      - name: rsync
+        run: |
+          set -ex
+          #
+          # 1. sync everything except `data/blog` (without --delete)
+          # 2. sync `data/blog` WITH --delete
+          #
+          # 1.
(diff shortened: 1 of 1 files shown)


timlrx:
This is too smart 😄

Now I really feel like creating a documentation template that expands on this concept. I think we should automatically disable the newsletter box since it is the only feature in the static site that cannot actually be used. 

Another consideration is maintenance - I don't want to break everyone's blog with an update. Having the deployable github pages is a start, but I can't guarantee that the settings and data are always backwards compatible, not to mention certain dependencies with breaking changes. Perhaps we should pin this to a version and let the user manually update it, instead of using main?

abernier:
about maintenance considerations, the `uses:` value has a ref `@v2` which will point to the workflow file of that branch

so no risk to break things

abernier:
> Another consideration is maintenance - I don't want to break everyone's blog with an update. Having the deployable github pages is a start, but I can't guarantee that the settings and data are always backwards compatible, not to mention certain dependencies with breaking changes. Perhaps we should pin this to a version and let the user manually update it, instead of using main?

For this, I've set a `ref` input, that defaults to `@v2` so the checkouted `timlrx/tailwind-nextjs-starter-blog` version is v2 as of now.

https://github.com/timlrx/tailwind-nextjs-starter-blog/blob/5aeedd58ad43c2ce226afeb40e21b43cfbe02c5b/.github/workflows/action.yml#L4-L22

abernier:
finally, once we have gh-page workflow, I'm not so sure it's a good idea compared to a simple fork, that stay "syncable"...

timlrx:
> finally, now that we have gh-page workflow, I'm not so sure it's a good idea compared to a simple fork, that stays "syncable"...

I think it's a interesting feature, probably more suited for someone who just wants to use the default config with minimal changes. Though if you are using a next.js blog, you are probably looking to customize it.

PR #969, opened 2024-07-14
Update contentlayer.config.ts

emmanuelh-dev:
To use the TOC (Table of Contents) with the default string, you might encounter an error due to incompatibility with the TOC properties. When attempting to parse it, it may break because it is already a string. However, with these changes, it works fine:
image

### DIFF
--- a/contentlayer.config.ts
+++ b/contentlayer.config.ts
@@ -43,7 +43,7 @@ const icon = fromHtmlIsomorphic(
 const computedFields: ComputedFields = {
   readingTime: { type: 'json', resolve: (doc) => readingTime(doc.body.raw) },
   slug: {
-    type: 'string',
+    type: 'json',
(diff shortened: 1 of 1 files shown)


timlrx:
Hi @emmanuelh-dev, thanks for the PR. Could you give an example where it might break if using a string rather than json?

emmanuelh-dev:
greatting tim,
Sure, i add the TOC to every page, wiout a component
image
but this broke becase when you do the build, the

timlrx:
Sorry for the late reply, but I am still confused over here. In the PR you changed the type of `slug` from `string` to `json` but it appears you are having issues with `toc`? From what I checked, the types on `slug` should be and is correctly set to `string` while the type on `toc` is also correctly set as `json`.

emmanuelh-dev:
Hi Tim, is just to avoid this

PR #978, opened 2024-07-20
Fix hydration warning on blog timestamp

krispya:
The timestamp between the client and Vercel server can be different giving a hydration warning. This suppresses the warning for the timestamp.


timlrx:
Thanks for the fix

PR #975, opened 2024-07-19
fix: resolve console warning when switching between light and dark modes

asanchezyali:
This commit addresses the issue where an error was logged in the console when the user switched between light and dark modes. The warning was caused by the inappropriate use of SVG tags in JSX. The solution involved correcting the SVG tag usage to comply with JSX syntax requirements.

Closes #974

PR #979, opened 2024-07-20
Fix Monitor SVG using snake case causing React warnings

krispya:
Fix Monitor SVG using snake case causing React warnings. That simple.


timlrx:
Fixed in #975

PR #985, opened 2024-07-28
more BASE_PATH catches

abernier:
this time for siteMetadata's `siteLogo` and `socialBanner`


abernier:
@timlrx and this one, lgty ?

PR #1005, opened 2024-08-20
ReactElement is better type than ReactNode as it has the generic props

zriyansh:
more here - https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement


timlrx:
Please try running the modified code to ensure it runs before submitting a PR

PR #1006, opened 2024-08-22
Removed class `max-h-screen` from `ListLayoutWithTags`

thekoc:
The tag items would overflow their container if there are too many of them. Removed the `max-h-screen` class to prevent the overflow.

CleanShot-2024-08-22@15 27 14@2x


ray-android:
I’m not sure why you’re experiencing overflow issues, but in my case, a scrollbar appears due to the overflow-auto setting.

PR #1012, opened 2024-08-27
Add decodeURI to ListLayoutWithTags

yanyao2333:
When the tag name is not in English, the original code will compare the UTF-8 encoded tag name with the original tag name, causing the tag to not be selected. This issue can be resolved by adding `decodeURI`.


timlrx:
Thanks, looks good to me!

PR #1022, opened 2024-09-08
Fix: syntax error

edenreich:
## Summary

Great template! thanks for creating this!

Commas are not needed here.

Prettier also corrected it for me when I simply save the file in VScode.


timlrx:
Thanks for the catch!

PR #1028, opened 2024-09-20
Allow Microsoft Clarity script in Content Security Policy

Violet-Bora-Lee:
- Add https://www.clarity.ms/ to script-src directive in CSP
- Resolve script loading error related to Clarity
image


timlrx:
Glad that you managed to get it working but in general I am not planning to add on the various analytics providers domain to the CSP whitelist - that should be handled by the individual

PR #1033, opened 2024-09-25
Add OpenSats to Examples V2

dergigi:
Not sure if you still accept new entries to the examples list, but [opensats.org](https://opensats.org/) is based on this project.

- https://opensats.org/blog
- https://github.com/OpenSats/website

Thanks for all the work you do!


timlrx:
Sure, happy to accept, and thanks for the work that you are doing as well!

PR #1055, opened 2024-11-05
Added support for youtube social icon below the profile image in the about page

Haidar0096:
Added support for youtube social icon below the profile image in the about page.

So now in the default author mdx file, you can put your youtube link, then run `yarn build`, and `yarn dev`, and it will show youtube icon in the about page below the profile image along with other social media icons.


timlrx:
@Haidar0096 thanks for the PR and glad that you manged to add a youtube icon to the blog, but I am not keen on accepting new PRs for social media sites currently - I think there are just too many to cover and it's easy enough to add a new one. Appreciate the effort though

PR #1058, opened 2024-11-07
Add bluesky

trillium:
# Overview

Add bluesky profile to app

Files changes:
- components/Footer.tsx
- components/social-icons/icons.tsx
- components/social-icons/index.tsx
- contentlayer.config.ts
- data/siteMetadata.js
- layouts/AuthorLayout.tsx

## Additional todos?
 - [ ] Add `discuss on bluesky` option in post layout


trillium:
Yeah, I hear that. I don't want to use X any longer personally, and would
like to provide a Bluesky option.

I don't intend to make further changes at this time, but I might loop back
later and create a ' share in Bluesky ' PR after this is merged

PR #1064, opened 2024-11-26
Header component update with dynamic navigation behavior

Erick-Leite:
This PR introduces significant updates to the Header component, to elevate the user experience to a new level. Now, there are three navigation bar options for the user to choose from: 'default', 'fixed', or 'dynamic'.


Erick-Leite:
Consider the last 5 commits

PR #1067, opened 2024-12-07
align beehiiv env variables with pliny implementation

wimdetroyer:
align beehiiv env variables in env.example file with pliny implementation

See: https://github.com/timlrx/pliny/blob/main/packages/pliny/src/newsletter/beehiiv.ts 

This makes the example fool proof :)


timlrx:
Good catch

PR #1071, opened 2024-12-10
add TOC side right

tqt97:


scripts/rss.mjs
```diff
@@ -3,7 +3,7 @@ import path from 'path'
 import { slug } from 'github-slugger'
 import { escape } from 'pliny/utils/htmlEscaper.js'
 import siteMetadata from '../data/siteMetadata.js'
-import tagData from '../app/tag-data.json' assert { type: 'json' }
+import tagData from '../app/tag-data.json' with { type: 'json' }
```
tqt97, line 6: It's still in the stage 3. Let's wait until the stage 4 will be reached.

tqt97:
closed

PR #1123, opened 2025-02-22
up next-themes for react 19

kyleholgate:
`next-themes` v0.3.0 is not compatible with react 19. Upping the version to v0.4.4 which is compatible.

image


timlrx:
Thanks, you should push the updated yarn lock file as well

timlrx:
Merge updated package in #1136

PR #1129, opened 2025-03-02
Error when creating new blog post

t14:
When creating a new blog post the following TypeError occurs 

```
Unhandled Runtime Error

[ Server ] Error: Cannot read properties of undefined (reading 'includes')

eval
./app/blog/%5B...slug%5D/page.tsx
Module.generateMetadata
./app/blog/%5B...slug%5D/page.tsx

```

Checking that `img` is defined before calling the the `includes` method resolves the issue


timlrx:
Thanks for the fix

PR #1132, opened 2025-03-10
style: remove content from code pseudo-elements

xianminx:
This PR updates the `inline code` style. The default tailwind typography will add backticks `` ` `` around the code.

* The old style: 
    image

* The updated style removes the surrounding backticks `` ` `` 
    image


timlrx:
I think the default look is fine, but feel free to update it on your own blog!

PR #1138, opened 2025-03-31
Fix for regex trialing slash if `trailingSlash: true`

### LINKED ISSUE #1137: Regex for removing trailing slash on list layouts not working
**Describe the bug**
If `trailingSlash: true` is defined in next.config.js, the current regex rule of `.replace(/\/page\/\d+$/, '')` should be changed to `.replace(/\/page\/\d+\/?$/, '')`.

**To Reproduce**

1. Set `trailingSlash: true` in `next.config.js,`
2. Go to http://localhost:3000/blog
3. Click on 'Next' and 'Previous' pagination links.
4. See error: You will see the trailing slash is not removed properly and now causing 404 issues.

**Expected behavior**
Trailing slash from regex not being removed

**Screenshots**

Image

**System Info (if dev / build issue):**

- OS: `Mac OS Sequoia 15.3.2`
- Node version: `v20.18.0`
- Npm version: `10.8.2`

**Browser Info (if display / formatting issue):**

- Device: Desktop
- Browser: Chrome 
- Version: 134.0.6998.166 (Official Build) (arm64)

ryanfitton:
Fixes #1137

This PR fixes the issue on the regex in list layouts for removing the trailing slash used in pagination.

If `trailingSlash: true` in `next.config.js`, this will now remove the trailing slashes.

Also removed some `console.log()` messages.

### DIFF
--- a/layouts/ListLayout.tsx
+++ b/layouts/ListLayout.tsx
@@ -26,7 +26,7 @@ function Pagination({ totalPages, currentPage }: PaginationProps) {
   const lastSegment = segments[segments.length - 1]
   const basePath = pathname
     .replace(/^\//, '') // Remove leading slash
-    .replace(/\/page\/\d+$/, '') // Remove any trailing /page
+    .replace(/\/page\/\d+\/?$/, '') // Remove any trailing /page and optional trailing slash
   const prevPage = currentPage - 1 > 0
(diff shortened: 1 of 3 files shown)


timlrx:
Good catch, thanks for the fix!

Issue #936, opened 2024-06-09
 Feature Suggestion: Table of Contents (TOC)


byungjin0826:
Hi timlrx, 

I hope this message finds you well. Firstly, I want to express my gratitude for developing and maintaining the package. It has been incredibly useful for creating and managing my blog, and I appreciate all the effort you’ve put into it.

I have a suggestion that I believe could enhance the usability and functionality of the package: the addition of a Table of Contents (TOC) feature.

Currently, the package provides excellent tools for blog creation and management, but I feel that adding a TOC feature could significantly improve the user experience. Here are a few reasons why:

- Improved Navigation: A TOC would allow readers to easily navigate through longer blog posts, quickly jumping to sections of interest without scrolling.
- Better Content Organization: It helps in organizing the content more effectively, making it more readable and user-friendly.
- Enhanced SEO: Search engines often use TOCs to understand the structure of a page, potentially improving the blog’s SEO performance.

If this feature could automatically generate a TOC based on the headers within the post, it would greatly streamline the process for users.

Additionally, if this feature is already available within the package and I might have missed it, could you please guide me on how to implement or enable it?

Please let me know if you need any further details or if there’s any way I can assist in implementing this feature. I’d be happy to contribute in any way possible.

Thank you for considering my suggestion. I look forward to any updates or feedback you might have.

timlrx:
It already exists, but it's an opt-in feature for each post so you can insert it where it fits best. Check out https://tailwind-nextjs-starter-blog.vercel.app/blog/new-features-in-v1#table-of-contents-component.

Just include the following component in your post: `<TOCInline toc={props.toc} />`. You can see an example in the [raw markdown file](https://raw.githubusercontent.com/timlrx/tailwind-nextjs-starter-blog/main/data/blog/release-of-tailwind-nextjs-starter-blog-v2.0.mdx) as well.

PxlSyl:
Hi, yes it already exists. I recently implemented the toc feature as an automatic sidebar when navigating to the posts (I may modify it to make it optional) on my i18n version, the reason I did this is for long blog posts and accessibility: I think it's better to have a floating buttton that can trigger the toc component, than the TOC at the top of the posts. If @timlrx is okay, I may do some PR for some of the features I implemented :)

timlrx:
[...]

PR #1141, opened 2025-04-07
feat: add sticky TOC to post layout

Misaki-Akeno:
Add table of contents to the post sidebar, as implication of #936.

Improves navigation experience for longer articles.


Misaki-Akeno:
not sticky but normal TOC😀

PR #1157, opened 2025-06-13
docs: added MindMire to showcase – supports fully interactive JSX blog posts

khokonm:
Hey! 👋
I’m adding [MindMire.com](https://mindmire.com/) to the showcase list — my personal blog where I offload raw, unfiltered, and unstructured thoughts. Think of it as a digital mind-dump.

I’ve also modified the template to support JSX-based interactive posts, allowing me to experiment with dynamic content inside blog entries — like live components, inputs, toggles, and more.

Big thanks for this amazing starter blog. It’s the perfect foundation for creative experiments like this!


timlrx:
Thanks, looks good but can you add it to the end of the list? First come first serve principle :)

khokonm:
Ah yes, sure. No problem at all!

PR #1170, opened 2025-07-21
A bug occurs when a tag is registered in Korean.

woongsik2:
This is the fix related to issues that occur when using Korean characters as tag names.

1. Fixed the issue where posts were not displayed properly when selecting a tag written in Korean.
2. Fixed the issue where the browser title appeared as encoded text when selecting a Korean tag.
3. Fixed the issue where the header title appeared as encoded text in mobile browsers when selecting a Korean tag.

These issues do not occur when using English tags, but when using Korean tags, the screen appears as shown in the image below.

> Browser title
image

<br/><br/>

> The difference between tapping a Korean tag and an English tag
image
image

<br/><br/>

> When a tag is tapped in a mobile browser
image

### DIFF
--- a/app/tags/[tag]/page.tsx
+++ b/app/tags/[tag]/page.tsx
@@ -37,7 +37,7 @@ export const generateStaticParams = async () => {
 export default async function TagPage(props: { params: Promise<{ tag: string }> }) {
   const params = await props.params
   const tag = decodeURI(params.tag)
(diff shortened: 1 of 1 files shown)


WikiNeo:
@woongsik2 Thanks, I have tested it and it works well for both tag page and tag page pagination after deployment.

PR #1221, opened 2026-03-08
Add rlodhiya.dev to list of content-collections examples

rldhy:
## Changes

- Add new section for examples that have been implemented with content-collections
- Add https://rlodhiya.dev to list of examples

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ yarn-error.log*
 
 # Contentlayer
 .contentlayer
+
+# Jetbrains IDEs
+.idea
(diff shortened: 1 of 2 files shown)


rldhy:
Hey @timlrx and other maintainers! 

Would love to get your opinion(s) on the idea of potentially migrating this repository from using contentlayer to something like content-collections. I have been using this template, and tried the migration on my own repo as an example. 

If you believe it will be worthwhile for others, I am happy to create a V3 for this template, or a new forked template based on this that enables using tooling that is better maintained for content management.

